Issue #32, opened 2016-07-05
create use_first_valid_of function


sfirke:
For saying, use A if not NA, otherwise use B if not NA, otherwise... could just be a big case_when call (**edit:** case_when is probably not right given variable # of args, for loop may be simplest).  Optionally create a new variable indicating which one was used.

Besides being more readable than nested ifelse, it would work for dates, which is vexing af if you run into this: `http://stackoverflow.com/questions/6668963/how-to-prevent-ifelse-from-turning-date-objects-into-numeric-objects`

In fact, have it print a warning if you feed it at least one input vector with class Date, suggesting that they specify `force_class = "date`.

sfirke:
Like this code I just reviewed:

currentroster <- currentroster %>%
  mutate(recent_rating = ifelse(!is.na(rating2016), rating2016, 
                          ifelse(!is.na(rating2015), rating2015,
                                               rating2014)))

sfirke:
Numeric, logical, character instances are easy.  Is it important to retain factor class if all inputs are factors?  And how should this work for dates?

Maybe allow the user to specify a class and then the empty result vector will be initialized with that type of NA?

sfirke:
Not going to have this create a new variable indicating the source vector used for each value.  That's a non-critical option, and this is clearest if it always returns a single vector.

PR #45, opened 2016-07-24
add use_first_valid_of() function

sfirke:
has test coverage + adds to the vignette


sfirke:
closes https://github.com/sfirke/janitor/issues/32

Issue #75, opened 2016-10-22
Proposal for dups_ family of functions


rgknight:
It appears that janitor is evolving to have a family of duplicate-handling functions (#72 , #18).

I would like to propose that we adopt a common naming convention of `dup_` for this family of functions. 

The core functions that we could identify right now would be:

- `dups_check` Check whether there are duplicates and handle. Should be pipe-able.
- `dups_count` Return a count of duplicates
- `dups_get` A simpler and faster (#67) version of the current `get_dupes` that eliminates the counting, leaving it up to the user whether she wants to merge in dups_count or run `count()` themselves
- `dups_fuzzy` or `dups_possible` Implementation of #72 

This does read a little more awkward to me than the reverse (`check_dups`, `count_dups`, etc), but it would allow for users to see all available duplicate handling functions by typing "dup_". 

The workflow would be:
- Litter your code with defensive `dups_check` assertions to make sure that no duplicates are being introduced.
- If  `dups_check` fails use `dups_count` and/or `dups_get` interactively to problem solve

A basic implementation could be:

```
dups_get <- function(df, ...){
  vars <- lazyeval::lazy_dots(...)
  target <- df %>% select_(.dots = vars)
  dup_index <- duplicated(target) | duplicated(target, fromLast = TRUE)

  df[dup_index, ]
}
```

```
dups_count <- function(df, ..., sort = FALSE){
  vars <- lazyeval::lazy_dots(...)

  dplyr::count_(df, vars, sort = sort) %>%
    dplyr::filter(n>1) %>%
    dplyr::rename(.duplicates=n)
}
```

```
dups_check <- function(df, ..., level="warn"){

  levels <- c("warn", "stop", "message", "logical")

  if (!(level %in% levels)){
    stop(sprintf("level must be on of %s", paste0(levels, collapse = ", ")))
  }

  vars <- lazyeval::lazy_dots(...)
  target <- df %>% select_(.dots=vars)

  num_dups <- sum(duplicated(target))

  dup_message <- sprintf("There are %d duplicates", num_dups) # would be nice to parse the df name instead of gettting "." when piped

  if (level == "logical"){
    return(num_dups != 0) # Logical returns true/false only
  }
  else if (num_dups == 0){
    invisible(df)
  }
  else if (level == "message"){
    message(dup_message) # Would be nice to return the call as well
  }
  else if (level == "warn"){
    warning(dup_message)
  }
  else if (level == "stop"){
    stop(dup_message)
  }
[...]

Issue #74, opened 2016-10-21
How is use_first_valid_of different from dplyr's coalesce?


rgknight:
Would be helpful to have an explanation in the help file, if it is indeed different.

https://github.com/hadley/dplyr/blob/master/R/coalesce.R

sfirke:
Great catch.  It's not that different :-(  I didn't know about `dplyr::coalesce`.

I compared them and think they are close enough that `use_first_valid_of` should be deprecated and phased out, with a message to use `dplyr::coalesce` instead.  There are two features in `use_first_valid_of` that are missing in `coalesce`:

1. If there are any type mismatches in the input vectors, the output is coerced to a character vector (with a message).  In `coalesce`, the operation fails with a warning.  Stylistic difference, and okay with me.
2. Factors are coerced to `character` in `use_first_valid_of`, while `coalesce` fails:
```
> x <- factor(c("a", NA))
> y <- factor(c("b", "c"))
> use_first_valid_of(x, y)
[1] "a" "c"
Warning message:
In use_first_valid_of(x, y) :
  Input vectors are of class 'factor' and will be coerced to class 'character' before combining
> coalesce(x, y)
[1] a    <NA>
Levels: a
Warning message:
In `[<-.factor`(`*tmp*`, i, value = 2L) :
  invalid factor level, NA generated
```

Course of action:

- [ ] Deprecate `use_first_valid_of` with message referring to `coalesce`
- [ ] Post dplyr issue identifying issue 2 above; acceptable if `coalesce` can't handle factors but in that case it should fail with an error message, not return the current output, IMO.

rgknight:
I haven't looked too closely at  `convert_to_NA`, but it may also be a candidate for deprecation with a pointer to dplyr's `na_if`?

rgknight:
I submitted an issue on dplyr for this https://github.com/hadley/dplyr/issues/2254

sfirke:
Updating the vignette to remove the deprecated `convert_to_NA` and I kind of want to put it back after reading this discussion: https://github.com/tidyverse/dplyr/issues/1972...

PR #80, opened 2016-11-28
Moving conversation forward

rgknight:
This should perhaps be several smaller PRs, as this PR includes several very different ideas.

- Enforce consistent naming conventions per https://github.com/sfirke/janitor/issues/75
  - Anything that is primarily intended to modify the output of `tabyl` or `crosstab` now has the prefix `adorn`.
  - The set of verbs is now `adorn`, `check`, `clean`, `convert`, `count`, `get` and `remove`, in addition to the core `tabyl` and `crosstab`. This is probably still too many, but I intend this more to start the conversation than to end it. 
- Break `get_dupes` out into a family of functions, per https://github.com/sfirke/janitor/issues/75
  - The existing version of `get_dupes` does several things. This PR simplifies `get_dupes` to do only one thing: get duplicates. You can now separately `check_dupes` and `count_dupes`. 
- Eliminate `use_first_valid_of` and `convert_to_na` per https://github.com/sfirke/janitor/issues/74
- Add `check_encoding`, which I use to help make sure that encodings don't change when I am working with non-UTF files (e.g., someone opens the csv file and re-saves it using Excel, introducing a BOM, which readr doesn't currently handle well).
  - The point of adding this is more to demonstrate how `check` function might develop, since they are such a critical step in data cleaning

Let me know if there is anything I can to do make this easier to work with.


sfirke:
@rgknight what do you think about splitting this into multiple smaller PRs, based on my feedback above?

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: janitor
 Title: Simple Tools for Examining and Cleaning Dirty Data
-Version: 0.2.1.9000
+Version: 0.3
```
rgknight: I think that convention is to bump the version now so that we can tell the difference between which of our code is using the prod version and which is using this version. I agree that a PR is not the way to do this -- you should probably commit directly. People will know it's a dev version because it's installed via install_github instead of cran. Alternatively, you could create a "v0.3" branch that I can submit PRs to and then we can pull that branch into master when you are satisfied. 

Can you do one of these soon as it will impact future PRs?

janitor.Rproj
```diff
@@ -0,0 +1,17 @@
+Version: 1.0
```
rgknight: Oh, really? You should definitely start using projects for all of your work. If you are using RStudio, it should be a standard part of your workflow. Basic intro here https://support.rstudio.com/hc/en-us/articles/200526207-Using-Projects

Could you accept https://github.com/sfirke/janitor/pull/79? This will make development easier for me.

DESCRIPTION
rgknight: Never mind! Ignore my previous comment. I now think it's fine to keep the dev version at 0.2.1.9. Sorry for confusion.

janitor.Rproj
sfirke: I accepted #79

PR #82, opened 2016-12-01
Deprecate per #74

rgknight:
Deprecate use_first_valid_of and convert_to_NA, using .Deprecated and plyr::liply as a guide.


rgknight:
Closes #74.

It looks like my `# EXCLUDE COVERAGE START` was wishful thinking. Do you have a resource on how to ignore lines / files from Codecov?

sfirke:
I don't know a bypass for Codecov, but I think we can keep the tests.  They generate warnings, but it passes the Travis CI build and as long as the deprecated functions can be run I think they should have tests - what if some other change would cause them to fail?

This looks good.  I am a little conflicted about dropping `convert_to_NA` since it takes multiple arguments to convert to NA and it works on a data.frame, both of which are more tedious with `na_if`.  But, it also only works for string variables.  And I think the package is leaner and more focused without it.  So let's deprecate - maybe it has a fan and they can register a complaint.

Could you grant me access to commit to this PR?  I made a few small modifications but can't commit to it.  See https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/.

rgknight:
Thanks. I granted commit access. 

The tidyversian way to recode multiple columns with `na_if` would be using one of the `mutate_if` /  `mutate_each` functions, which is pretty straightforward, though I am not a big fan of the `vars` and `funs` syntax. 
```
iris %>% mutate_all( funs( na_if( ., 1.4))
```

PR #99, opened 2017-03-22
bring add_totals functions up to v1.0 as add_totals

### LINKED ISSUE #57: adorn_crosstab's helper functions always skip the first column
`add_totals_row`, `add_totals_col`, and `ns_to_percents` all have it hard-coded that the first column is a character vector to be ignored.  Fine when they were non-exported helper functions for `adorn_crosstab`, but if exported, this should be more flexible.  There could be multiple non-numeric columns, in any order, or none.

Default to skipping an arbitrarily-wide number of initial non-numeric columns?  Or detect non-numeric columns, and exclude (for summing) or leave as is (for in-place conversion to percentages)?

sfirke:
Closes #57, lots of discussion there.


sfirke:
hmph I see I need to add a little test coverage.  Will do that.  @rgknight would you be up for reviewing the code in this PR?

R/ns_to_percents.R
```diff
@@ -18,32 +19,32 @@
 #'   # when total_n is needed
 #'   mtcars %>%
 #'   crosstab(am, cyl) %>%
-#'   add_totals_row() %>% # add a totals row that should not be included in the denominator
+#'   add_totals("row") %>% # add a totals row that should not be included in the denominator
 #'   ns_to_percents(denom = "all", total_n = nrow(mtcars)) # specify correct denominator
   
 ns_to_percents <- function(dat, denom = "row", na.rm = TRUE, total_n = NULL){
```
rgknight: One option for a v1.0 approach: 

This is a function primarily intended to modify a tabyl, so it would be an `adorn` function, probably `adorn_percent(denom="row")` for the percent of total over rows. 

If you included both `adorn_N` and `adorn_percent` in the pipe after a `tabyl`, it would behave like `adorn_crosstab` does now, or maybe just add columns for both if that's easier. 

It sounds hard to make `tably` create an environment that could allow for multiple `adorn_` functions to be stacked. I think it's possible, but I have no idea how to do it.

R/add_totals.R
```diff
@@ -1,9 +1,67 @@
...
+#' @return Returns a data.frame augmented with a totals row, column, or both.
+#' @export
+#' @examples
+#' library(dplyr) # for the %>% pipe
+#' mtcars %>%
+#'   crosstab(am, cyl) %>%
+#'   add_totals()
+
+
+add_totals <- function(dat, which = c("row", "col"), fill = "-", na.rm = TRUE){
```
rgknight: I prefer `adorn_totals` as this is something that is primarily intended to modify a `tabyl`. I think that `adorn_crosstab` may eventually become the outlier in the `adorn_` approach. I think I've given this feedback already, though, so I assume that you have heard it and decided on verb_object approach to things that modify a tabyl.

R/add_totals.R
```diff
@@ -1,9 +1,67 @@
...
+    dat$Total <- row_totals$Total
+  }
+  
+  dat
+}
+
+### Deprecated functions -----------------------------
 #' @title Append a totals row to a data.frame.
 #'
 #' @description
```
sfirke: Yep, good call. I'll add that.

- [ ] Deprecate the retired `add_totals_*` functions in their documentation

R/add_totals.R
sfirke: I think there's something to this, and I agree that `adorn_crosstab` will become the outlier (a) because the nomenclature of "crosstab" may go away and (b) if we can fully replicate its functionality with modular pieces like this.  In that case, I do like the common prefix of `adorn_` and could see this as `adorn_totals`.

I want to merge this in now rather than wait for us to figure out #101, since this closes existing bugs.  I'm not sure if it should be `add_totals` when we might later re-route people to `adorn_totals`... maybe `adorn_totals` now.  I think this won't be set in stone until we resolve #101, the big question, and so no matter what we call it now, the name and behavior of the totals function could ultimately change.  Say, if the totals function needs to add an attribute to a tabyl to convey info to `ns_to_percents` down the line.

sfirke: okay, I renamed it `adorn_totals`.

R/add_totals.R
```diff
@@ -1,9 +1,67 @@
...
+    
+    col_totals[nrow(col_totals), 1] <- "Total" # replace final row, first column with "Total"
+    dat <- dplyr::bind_rows(dat %>%
+                              stats::setNames(names(dat)) %>%
+                              dplyr::mutate_at(1, as.character), col_totals)
+  }
+  
+  if("col" %in% which){
+    # Add totals col
+    clean_dat <- clean_names(dat) # bad names will make select_if choke
```
sfirke: Done.

PR #105, opened 2017-03-30
closes #95 among other discussion

### LINKED ISSUE #95: checking that a join didn't alter the number of records
This would be a defensive `check_` for assertive programming.  There are two common scenarios where I want the result of a `_join()` to have the same number of rows as the `x` table:

1. `left_join` - if there are duplicate values in the right table, it'll increase the number of rows in the left
2. `inner_join` - if all values in the left table (order doesn't matter, for this to be sensible the user should specify the larger table as the left one) don't have a match in the right, the number of rows will decrease

Is that in scope for this package?  It is a common data cleaning task.

rgknight:
So here's the deal. 

assertive is good but not pipe-able (yet)

assertr is pipe-able but I find it very hard to access for new users and their default assert_unique tests each variable independly instead of jointly. 

I quite like checkmate from an initial investigation but it doesn't have duplicate checking (yet). Also not nicely pipe-able.

This code isn't very clean. I don't know enough about environments to figure out how to not duplicate everything. 

We could: 
1. Say it's not worth doing this because there are too many other options
2. Wrap assertive instead of generating our own messages
3. Just live with this messy code because its a small thing that may go away eventually

Also, advice on readr example? Don't want to import it just for the example, but wanted to make it realistic.


rgknight:
Hey Sam,

Are you interested in this? If not I will close and load it in a personal project. 

- Ryan

sfirke:
Hey!  Sorry to have been MIA.

I am unsure whether it belongs with janitor or whether it is better as a small separate packages of checks.  I lean toward the latter.  And I feel low-bandwidth right now and am trying to wrap up overhauling the tabyl() function, which is 75% done.  I don't want to hold you up if you feel passionately about this.

So maybe put it in a separate package?  That feels more in line with shifts I see Hadley making, going toward more smaller packages.  If I'm wrong and it obviously belongs w/ janitor, then we can merge them later.

When this is its own thing, let me know and I'll use it!  Also I know my colleague @adamMaier is interested in a similar topic; FYI, Adam.

sfirke:
Also w/r/t loading the example files above with readr, now dplyr 0.7.1 has some built in data sets to demonstrate joins, so you can use those for examples.

rgknight:
Ok, I put it up [here](https://github.com/rgknight/knightr) for now so I can access it on some projects. 

I don't know that I want to maintain a checking-specific package, since there are so many other checking-specific packages around. I see these functions as more of an introduction to assertions / checking for folks who are just starting to get serious about R, which could fit under the janitor mission, so let me know if you change your mind!

sfirke:
Tidying up this repo; I think the old discussion about (2017!!) is right, that this would be better off in a small separate package.  So I will close this PR.  I still see its usefulness.  Though I'm not sure if the landscape of assertive packages has changed in the meantime.

It's been a while, hope you are well Ryan!

Issue #130, opened 2017-08-03
minor improvements to clean_names


sfirke:
use `vapply` instead of `sapply`, insert `_` around `"percent"` after it's converted from the % sign.  Per discussion in #96, I'm breaking this off into its own small issue.

sfirke:
Completed by @Tazinho .

Issue #96, opened 2017-03-13
offer other case options as an argument in clean_names


sfirke:
Suggested by @masalmon [on Twitter](https://twitter.com/ma_salmon/status/841267599749062656).  I only use snake_case, but it would be more inclusive of other coding styles to offer camelCase or maybe PascalCase.  Perhaps wrap @Tazinho's [snakecase ](https://github.com/Tazinho/snakecase)package?

masalmon:
For the record, I only use snake case but I can imagine other people using `janitor` and wishing that, not sure they would install a package just for changing column names (not undermining your great work @Tazinho !) so that it'd be easier for them to have it inside `janitor`?

Tazinho:
Just some notes on that. It is super easy to get pascal- and camel-case once you, @sfirke  have converted to clean snake_case. So it might still be useful to build it up on the already available clean_names() function, which has all the tests for valid data.frame names. Of course you can wrap snakecase, just let me know any requirementes. Also note that snakecase currently imports purrr, stringr and magrittr-pipes, but I could translate to a baseR version if this is an issue.

On the long run I think it would be best if the functionality would be implemented in stringr/stringi. However, at the moment it is not, which is as far as I know, because these packages claim more generality and it is not always obvious what the snake case result should look like.

Tazinho:
Just wanted to let you know that snakecase is on cran now and it is planned to support a lot of stuff that might be interesting for you. See the [issues](https://github.com/Tazinho/snakecase/issues) for that or this talk:
https://cran.r-project.org/web/packages/snakecase/index.html
Note also that there are (currently experimantally) options to fill empty names and make sure that names are not dublicated.
However, I am not sure if it is a good idea to reproduce exactly your testcases in the long run. If you have any further ideas what should be supported for your usecase, don't hesitate to let me know.

maelle:
Nice work! Since I have a special character in my name I like the related issue 😀

sfirke:
@Tazinho congrats on the package being on CRAN!  I have a data source at work that gives me camelCase and it looks like `to_snake_case` will get it more to my liking 👏 

I think this would be nice to offer camelCase and PascalCase as options to `clean_names`, maybe a `case = ` parameter.  And I agree, it appears that once the current `clean_names` runs, the conversions with `snakecase` functions will be simple one-liners.

I like minimizing dependencies to the extent possible, but I don't think it'll be a dealbreaker here.
[...]

PR #131, opened 2017-08-16
Integrate snakecase-pkg into janitor

Tazinho:
As mentioned in #96 and #130 For the latter I changed "%" into ".percent_" instead of `"_percent_"`, as commented in the code, since starting with "_" would have been converted to "x_percent" by `make.names()` .

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -21,7 +21,7 @@ Depends:
 Imports:
     dplyr (>= 0.5.0),
     tidyr,
-    stringi
+    snakecase
 License: MIT + file LICENSE
 LazyData: true
 RoxygenNote: 6.0.1
--- a/R/clean_names.R
+++ b/R/clean_names.R
@@ -1,12 +1,31 @@
 #' @title Cleans names of a data.frame.
 #'
 #' @description
-#' Resulting names are unique and consist only of the \code{_} character, lowercase letters, and numbers.
+#' Resulting names are unique and consist only of the \code{_} character, numbers and letters regarding the specified \code{case}.
 #' 
 #' Accented characters are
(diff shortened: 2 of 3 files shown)


sfirke:
Looks good.  I have two questions:

1) Test coverage: I'd like there to be tests for each of the `case` options.  Are there similar tests from `snakecase` that we could quickly incorporate?  It would be a bit repetitive but we could start by just taking the existing set of `clean_names` tests and test it under each value of case, that would be quick.

2) The part in the documentation where it notes '_There are three "special" cases available_' - do you think those will be appealing to users in this use case, automatically reformatting their variable names?  I lean toward omitting that part for brevity.

Thanks Malte!

Tazinho:
Hi Sam,

1. There are many (more or less systematic) testcases in tests-to_any_case.R. I included your testcases for the case "snake" under the name janitor. I am currently on vacation, but afterwards will update these tests for the other cases. This should be straightforward and there should be no surprises. I wasnt sure if yould like to see the tests on my side, but I can include more tests and I'd be happy to offer support for any issues dealing with cases.

2. Hard to tell, but from a typical janitor usecase I'd expext that only snakecase, Upper/lower camel and all caps matter. Maybe parsed and mixed case also. I'd like to leave it up to you to edit the description.

Best, Malte

Tazinho:
Testcases for all other cases are now included within 
https://github.com/Tazinho/snakecase/blob/master/tests/testthat/test-to_any_case.R

sfirke:
Thanks for adding the additional test cases!  I'll tinker with the description per your comment above, re-read the code, hopefully this week, and then merge it.

sfirke:
Okay if I copy those tests to janitor?  Seems like they should call `clean_names()` directly in case it changes in the future.

Tazinho:
Definitely ok.

Issue #135, opened 2017-09-25
different default updates to clean_names


amartinNPS:
👋 

Very cool to see the `snakecase` integration!

I would like to suggest that clean_names have a different `case` argument by default.  While I like snakecase, I _believe_ that old janitor `cleannames()` produced something closer to lowercamel.

I have a bunch of code like

```
      test_stateid = sid_check_stateid(stateidentificationnumber),
      test_firstname = sid_check_firstname(firstname),
      test_lastname = sid_check_lastname(lastname),
```

that just broke because I updated janitor, and output of clean_names now defaults to `state_identification_number`.

guessing that anyone else with `clean_names` in their pipeline would see the same thing; in this case I would argue that the default behavior should prioritize consistency with prior releases even if there is a preference for snakecase on the development team here.

thoughts?  If there is agreement I will submit the PR!

sfirke:
I'm glad you filed this issue - I did not expect/intend for this to be a breaking change.  What is the input string that went from `stateidentificationnumber` to `state_identification_number`?  The prior version of janitor would turn `"a test variable name"` into `"a_test_variable_name"`.

almartin82:
I (believe) that the prior version of janitor turned
`LocalIdentificationNumber` (in the csv) into `localidentificationnumber`
(in R).

Now I am seeing `LocalIdentificationNumber` become
`local_identification_number` - and the only thing that happened (again, to
the best that I can determine) is that I updated janitor.

sfirke:
Ah okay, I can reproduce that as the fault of the janitor update.  Sorry for the breakage.  Camel case variable names in the raw data are now detected and that spacing is preserved via underscores.

This is tricky, since the old `clean_names` behavior doesn't map neatly to any of the new cases.  Before, `"LocalNumber"` became `"localnumber"` and `"Local Number"` became `"local_number"`; under the new options, these would both map to `"local_number"` (`case = "snake"`) or `"localNumber"` (`case = "small_camel"`).

I think this consistency is an improvement in janitor ... but I see how this will be a massively breaking change.  I will ponder that tradeoff and welcome feedback from anyone interested.

One option would be to offer another case option, `"old_janitor"`; then to deal with the breakage, you could find and replace `clean_names()` with `clean_names(case ="old_janitor")` instead of having to manually address the changed variable names.  Still a breaking change, but maybe this softens the impact?

rgknight:
I like the new behavior as the going forward option, but also feel bad
[...]

PR #139, opened 2017-10-11
Update clean_names.R

Tazinho:
Quick fix, to provide support for (exactly) the old behaviour (regarding #135).

### DIFF
--- a/R/clean_names.R
+++ b/R/clean_names.R
@@ -16,7 +16,7 @@
 #'  \item{ALL_CAPS: \code{"screaming_snake"} or \code{"all_caps"}}
 #'  \item{lowerUPPER: \code{"lower_upper"}}
 #'  \item{UPPERlower: \code{"upper_lower"}}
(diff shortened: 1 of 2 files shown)


sfirke:
Thanks for getting this rolling!  I think it needs tests to cover the new `"old_janitor"` argument - they can be the original test suite given that it's the original function 😝 - and probably every set of case tests should have a new test with a starting column name like `testCamelCase` since that's the breaking change we're addressing (and that I wish I'd included from the beginning).

sfirke:
I tried to commit the tests so I could merge this PR, but it looks like you deleted your fork of janitor, incl. the branch you started and I was working on above (`patch-3`).  Since the existing tests are passing, I'll merge the PR and commit the tests to master.  Thanks again!

Tazinho:
Yep, I forgot about that. Cool that u added the tests and sorry for the confusion!

Issue #77, opened 2016-10-26
missing column names after crosstab?


randomgambit:
Hello there,

First of all, congrats for this package. Coming from `stata` and `pandas`, this is EXACLY what `dplyr` needed. A r**obust, easy to use** package that excels at computing **tabulations and cross-tabulations**!

I have a quick question. Consider this simple example

```

var1<- c(5,10,2,2,NA)
var2<- c(NA,NA,NA,NA,NA)
var3<- c(5,10,2,2,1)
var4<- c('A','B','C','C','D')
var5<- c('G','B','C','F','D')
df = df=data.frame(var1,var2,var3,var4,var5)


> df
  var1 var2 var3 var4 var5
1    5   NA    5    A    G
2   10   NA   10    B    B
3    2   NA    2    C    C
4    2   NA    2    C    F
5   NA   NA    1    D    D
```

Now I compute a simple cross-tabulation

```
> df %>% crosstab(var1, var3,  percent = "row")
  var1 1 2 5 10
1    2 0 1 0  0
2    5 0 0 1  0
3   10 0 0 0  1
4   NA 1 0 0  0
```

I dont get why I only see the variable name `var1` above and not `var1` and `var3`. Given the table above, how could you possibly know that this is a cross-tabulation between `var1` and `var3`?
Am I missing something?

Thanks!

randomgambit:
by looking at the documentation, it seems this is the expected behavior of `crosstab`. 

In my opinion, there is room for improvement here. In essence, by not showing the second variable, we are losing information and our two-way frequency tables are not self-contained anymore. For a package specifically designed at cross-tabulations, I think this is a pretty serious issue.

What do you think?
In any case, your package is really great and I hope my comments will help you improve it!

sfirke:
I agree that the ideal result would contain information about both variables used in the crosstab and have actually put considerable thought into this.  I'm open to other ideas, but I haven't found a good way to do that without compromising the `data.frame` structure of the result.

For interactive use, something like that Stata result is nice, to have the full info.  But then the result can't be written to .csv, passed to `adorn_crosstab()`, joined to another data.frame, etc.

`gmodels::CrossTable` does something very similar to that Stata output: http://stackoverflow.com/questions/13043817/mimic-tabulate-command-from-stata-in-r  Perhaps that is your best bet.  Or from [this SO answer about xtabs](http://stackoverflow.com/a/2494278):
```
t1 <- rep(c("A","B","C"),5)
[...]

PR #145, opened 2017-10-31
implement adorn_col_name() and underlying attribute support

sfirke:
Have much of the underlying infrastructure here so far, need to finish drafting adorn_col_name() itself.


sfirke:
This will - finally - close #77.

PR #175, opened 2018-01-31
Merge hotfixes from 0.3.1 to CRAN into master

sfirke:
Only partially did this after creating 0.3.1 :grimacing:


sfirke:
In what critics are calling a stupid move, this PR now has just 1 change - to `cran_comments.md`.  Turns out I manually made the useful changes (code, documentation, NEWS) to the master branch already.  Lesson is that I should have done that via merging a branch for a better historical record / transparency, instead of duplicating commits.

sfirke:
Feels dumb to merge this and phony to delete it, so will merge it ... at least I can see what OSs I tested on for the 0.3.1 release :man_shrugging:

PR #146, opened 2017-11-08
increased tidyeval use

bfgray3:
* take advantage of tidyeval when possible instead of using underscore-postfixed tidyverse functions
* other small changes

PR #178, opened 2018-02-09
increase tidyeval usage

bfgray3:
this is a replacement for #146. sorry for the confusion in that pull request.

### DIFF
--- a/R/tabyl.R
+++ b/R/tabyl.R
@@ -172,9 +172,7 @@ tabyl_2way <- function(dat, var1, var2, show_na = TRUE, show_missing_levels = TR
 
   # Optionally expand missing factor levels.
   if(show_missing_levels){
-    combos <- tidyr::complete(tabl %>% dplyr::select(-n), !!!rlang::syms(names(tabl)[1:2])) # this is pretty ugly - using dplyr keeps col types the same making for easier join, vs. expand.grid
-                                                               # would be nice to just complete() tabl and skip the join, but couldn't get the eval to work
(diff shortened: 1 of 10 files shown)


sfirke:
Just taking a quick look, I think this is even better than the rebase.  It looks like you used the current dev version of these functions as the base for your changes so it should be easily mergeable.  

I'll try to review this within the next week.  I see changes like switching from deprecated `count_` to `count()` with tidyeval - that is great!

R/tabyl.R
```diff
@@ -172,9 +172,7 @@ tabyl_2way <- function(dat, var1, var2, show_na = TRUE, show_missing_levels = TR
 
   # Optionally expand missing factor levels.
   if(show_missing_levels){
-    combos <- tidyr::complete(tabl %>% dplyr::select(-n), !!!rlang::syms(names(tabl)[1:2])) # this is pretty ugly - using dplyr keeps col types the same making for easier join, vs. expand.grid
-                                                               # would be nice to just complete() tabl and skip the join, but couldn't get the eval to work
-    tabl <- suppressMessages(dplyr::full_join(tabl, combos))
+    tabl <- tidyr::complete(tabl, !! var1, !! var2)
```
sfirke: @bfgray3 I don't remember all of my struggles originally with this but my comment indicates that this was a kludge but the best I could do.  I seem to have succeeded in simplifying this as I'd originally hoped... it passes the tests so should be good but do you see any risks here?

bfgray3: Nice work! Definitely much better than the update I had made. :thumbsup:

Issue #50, opened 2016-08-05
function to compare data.frames that should be the same - variable names and col type consistency


almartin82:
here's my use case - I am writing functions to suck up _n_ state export files.

I can't depend on `read.csv`'s type hinting because I have situations where the _nth_ data file will contain a data type (say, 'K' for grade, where grade had always been integer on the first 10 files) that doesn't play nicely.  

To solve that, I'm reading the raw files in as character, and then doing type conversion myself.

But this seems like a janitor kind of job
![image](https://cloud.githubusercontent.com/assets/782135/17444927/98242458-5b10-11e6-971f-9da0d50de5e5.png)

Is this in scope/ out of scope?  Any thoughts about how to move forward?  @chrishaid would love your thoughts here as well

sfirke:
I have this problem too.  I got 20+ spreadsheets back that were supposed to use the same template, but some users had added or removed columns.  Combine that with type mismatches and it took a lot of work to bind them together.  I took the same approach as you, with reading them all as characters.  It was not good.

The new `readr::read_csv` 1.0.0 that Hadley dropped yesterday is some help, since you can quickly save what you determine to be the correct column specs and then pass that to all other `read_csv` calls.  But that doesn't work for rearranged or added/removed columns.

I've been thinking: a function that takes two data.frames and returns a description of how the 2nd is different than the 1st.  Like:
```
27 Identical columns
2 columns of same name, different type: year (date vs. string), ID (int vs. string)
1 absent column: rating (str)
3 additional columns: school1 (int), school1_name (str), district (str)
```
You'd use this interactively to quickly compare data.frames and operate on them such that they can be bound.

What do you think?  I'm not sure what format the result should be (showing both DFs vs. the leaner "how is df2 different than df1?", a data.frame vs. a list or text output).

sfirke:
My typical headache is two data.frames that are supposed to have the same columns and types.  I call `bind_rows` and it fails, telling me the first error.  Then I fix one thing at a time, not having a sense of how long this cycle will go.  Whipping up comparison function code on the fly is non-trivial.

almartin82:
Agreed, this describes a lot of common data import problems.

One potential end product would be a function that:

- takes a list of data frames
- figures out the minimum common type per column
- does the missing / additional column analysis
- has rule-based options (return all / return common) for rbinding it all together.

sfirke:
[...]

PR #179, opened 2018-02-10
compare_dfs

bfgray3:
I have attempted to take a first stab at including the function discussed in #50, based off your rough draft already there.  I didn't write any docs or tests yet since I wanted to get your thoughts first before I went ahead and did that.  This particular draft does introduce a (small) new dependency in `utils`, which I think should be attached by default most of the time anyway, but I'm sure there are other ways to do this.

### DIFF
--- a/R/compare_dfs.R
+++ b/R/compare_dfs.R
@@ -1,0 +1,53 @@
+check_bindable <- function(df1, df2){ # should eventually take a list of dfs
+  vars_in_df1_only <- names(df1)[!names(df1) %in% names(df2)]
+  vars_in_df2_only <- names(df2)[!names(df2) %in% names(df1)]
+  
+  # subset DFs to only the same for class investigation
+  common_vars <- Reduce(intersect, list(names(df1), names(df2)))
+  df1_common <- df1[, common_vars]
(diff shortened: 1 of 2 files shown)


R/compare_dfs.R
```diff
@@ -0,0 +1,70 @@
+get_first_col_classes <- function(dat) {
```
sfirke: Could this function simply be `vapply(dat class, character(1))`?

sfirke:
Thanks for taking a shot at this.  Let's discuss the desired behavior in #50 and then the details of the code implementation here?

R/compare_dfs.R
bfgray3: It turns out that `vapply` will error when the function `FUN` returns a longer-than-expected result for an element of the input  `X`:

```
> tt <- data.frame(x = 1, y = 2)
> class(tt$y) <- c("a", "b")
> lapply(tt, class)
$x
[1] "numeric"

$y
[1] "a" "b"

> vapply(tt, class, character(1))
Error in vapply(tt, class, character(1)) : values must be length 1,
 but FUN(X[[2]]) result is length 2
```

PR #181, opened 2018-02-11
changed link catalog of functions. 

kgilds:
Previous link was pointing to a 404 page. Updated link to https://sfirke.github.io/janitor/reference/index.html

## Description
changed link to reference page of package website.

## Related Issue
 Should and hopefully fixes Issue 23

### DIFF
--- a/README.Rmd
+++ b/README.Rmd
@@ -67,7 +67,7 @@ The current development version is significantly ahead of version 0.3.0 on CRAN 
 
 ## Using janitor
 
-Below are quick examples of how janitor tools are commonly used.  A full description of each function can be found in janitor's [catalog of functions](https://github.com/sfirke/janitor/blob/master/vignettes/introduction.md).  
+Below are quick examples of how janitor tools are commonly used.  A full description of each function can be found in janitor's [catalog of functions](https://sfirke.github.io/janitor/reference/index.html).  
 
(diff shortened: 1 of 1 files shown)


sfirke:
I see why you have this link point toward the list of functions on the packagedown site.  But it actually originally pointed to this vignette: https://github.com/sfirke/janitor/blob/master/vignettes/janitor.Rmd  Which used to be called "introduction" but got renamed to "janitor", thus the broken link.

With the packagedown-created list of functions, this vignette is less-good, but I think still worth having as it illustrates usage of the functions.  But I need to render that as a .Md so people can see the output of those calls.  Then this link can point there.

This is a busy week for me, so might be next week before I can generate that file + then this PR can point to it instead.  Sorry for the delay on such a minor-seeming thing.  Just writing now as to not leave you hanging, and to say thanks for the PR!

kgilds:
No worries, I was curios  if you would want it going to the the website. This was my first pull request and I learned a fair amount doing it. I love the package!

sfirke:
I'm honored that this was your first PR!  As soon as I get that .Md file updated for it to point to, I'll post here and then you can update your PR and we'll merge it.

sfirke:
I merged it!  Thank you Kevin for spotting this and setting up the fix :+1:

PR #187, opened 2018-03-08
Documentation update for remove_empty

garthtarr:
The documentation had singular "row" and "col" (which I actually prefer) whereas the accepted inputs are plural "rows" and "cols".

### DIFF
--- a/R/remove_empties.R
+++ b/R/remove_empties.R
@@ -4,7 +4,7 @@
 #' Removes all rows and/or columns from a data.frame that are composed entirely of \code{NA} values.
 #'
 #' @param dat the input data.frame.
-#' @param which one of "row", "col", or \code{c("row", "col")}
+#' @param which one of "rows", "cols", or \code{c("rows", "cols")}
 #' @return Returns the data.frame without its missing rows or columns.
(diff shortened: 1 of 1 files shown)


garthtarr:
Also, suggestion: default values for `which` should be set.  I'd encourage a `which = c("rows","cols")` as the default (maybe it's just me, but I almost always use `remove_empty_rows()` and `remove_empty_cols()` together.

sfirke:
I'll take a closer look tonight or tomorrow, but these sound like promising
improvements. Sneaking under the wire for version 1.0!

sfirke:
Good catch on this mismatch, thank you.  I think I like `"rows"` over `"row"` because it reads more like a sentence: "`remove_empty("rows")`.  But don't feel strongly, why do you prefer "row"?

As to the default argument, good question.  I will open an issue to discuss there, then this can just be merged.

garthtarr:
I also don't feel strongly about the plural vs singular, there's no need to change. I was thinking about functions like `ncol()` and `nrow()` where it's singular.  Mostly I'm lazy and favour typing less than more!

PR #214, opened 2018-06-24
Add time handling to excel_numeric_to_date

### LINKED ISSUE #210: Feature Request: Allow Times in excel_numeric_to_date
What do you think about making `excel_numeric_to_date` also handle times?

```r
excel_numeric_to_date <- function (date_num, date_system = "modern", include_time=FALSE) {
  if (!is.numeric(date_num)) {
    stop("argument `date_num` must be of class numeric")
  }
  ret <-
    if (date_system == "mac pre-2011") {
      as.Date(date_num, origin = "1904-01-01")
    } else if (date_system == "modern") {
      as.Date(date_num, origin = "1899-12-30")
    } else {
      stop("argument 'created' must be one of 'mac pre-2011' or 'modern'")
    }
  if (include_time) {
    ret <- as.POSIXlt(ret)
    total_sec <- 86400*(date_num %% 1)
    ret$sec <- total_sec %% 60
    ret$min <- floor(total_sec/60)
    ret$hour <- floor(total_sec/3600)
  }
  ret
}
```

billdenney:
Fix #210

A few notes:

I'm not sure if you want to describe the minor change of taking the floor of the value as a breaking change in the NEWS file.  It is very subtle and people shouldn't have been going into the date object that way.

While working up the test cases, I found a bug in base R (https://bugs.r-project.org/bugzilla3/show_bug.cgi?id=17437).  I worked around it in the result, and I think that the if the behavior changes, the code should be fine long-term.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,14 @@
 
 The new function `row_to_names` handles the case where a dirty data file is read in with its names stored as a row of the data.frame, rather than in the names.  This function sets the names of the data.frame to this row and optionally cleans up the rows above and including where the names were stored.  Thanks to **@billdenney** for writing this feature.
 
+## Minor features
+
+`excel_numeric_to_date()` can now provide times as well with the new `include_time` argument.
+
+## Breaking changes
+
+As part of `excel_numeric_to_date()` handling times, if a Date-only result is requested (the default) any fractional part of the date is now removed.  The date itself is identical, but the underlying object now has only the integer part of the date.
+
 # janitor 1.0.0  (2018-03-17)
 
 ## Release summary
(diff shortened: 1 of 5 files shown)


NEWS.md
```diff
@@ -4,6 +4,14 @@
 
 The new function `row_to_names` handles the case where a dirty data file is read in with its names stored as a row of the data.frame, rather than in the names.  This function sets the names of the data.frame to this row and optionally cleans up the rows above and including where the names were stored.  Thanks to **@billdenney** for writing this feature.
 
+## Minor features
+
+`excel_numeric_to_date()` can now provide times as well with the new `include_time` argument.
+
+## Breaking changes
+
+As part of `excel_numeric_to_date()` handling times, if a Date-only result is requested (the default) any fractional part of the date is now removed.  The date itself is identical, but the underlying object now has only the integer part of the date.
```
sfirke: This may be unclear to most readers (it was to me ).  I think the typical reader just wants to know, do I need to worry about this?  Perhaps an example of before and after objects returned from the same call will clarify.

sfirke:
Thanks again @billdenney !  Sorry for the delayed review + merge, I blame :baby:

billdenney:
No worries at all!  I have two kids and fully understand!

Would you like me to update the NEWS and vignette (I’m traveling, so it’d probably be about 2 weeks before I can get to it), or are you already in process for those?

sfirke:
I will take a crack at 'em!

PR #218, opened 2018-07-16
Correct second handling so that 60 seconds are not possible

### LINKED ISSUE #217: excel_numeric_to_date with include_time = TRUE can return 60 for seconds
Split off from #216, won't get to this in the next patch release that CRAN is requesting.

Here's what I have as I started to address that floating point imprecision can cause `x %% 60` to return 60, which shouldn't be possible.
```

    # Because of floating point imprecision, total_sec %% 60 can return 60 for ret$sec
    # E.g., 86400 * (42001.1 %% 1) = 8639.99999987427145 when it should be 8640, then it %% 60 = 60
    # Presumably ret$min could also be 60
    # The following code block addresses situations where seconds or minutes are 60
    # See https://github.com/sfirke/janitor/issues/216
    if(ret$sec == 60){
      ret$min <- ret$min + 1
      ret$sec <- ret$sec - 60
    }
    if(ret$min >= 60){
      ret$hour = ret$hour + 1
      ret$min = ret$min - 60
    }
...
```
But the cascading changes here are massive.  Say it was 23:59:60 on February 28th.  That should probably be represented as the identical 00:00:00 on March 1st (if not a leap year).  The method I have above of manually incrementing each unit is too tedious.

I welcome thoughts - does anyone have a solution to convert `07:00:60` to `07:01:00` in an elegant, accurate way?  And how if at all is it a problem to currently output `07:00:60`?

billdenney:
Fix #217

### DIFF
--- a/R/excel_dates.R
+++ b/R/excel_dates.R
@@ -34,10 +34,17 @@ excel_numeric_to_date <- function(date_num, date_system = "modern", include_time
   }
 
   # Manage floating point imprecision
-  date_num_days <- (date_num * 86400) %/% 86400
+  date_num_days <- (date_num * 86400L + 0.001) %/% 86400L
+  date_num_days_no_floating_correction <- date_num %/% 1
+  # If the day rolls over due to machine precision, then the seconds should be zero
+  mask_day_rollover <- date_num_days > date_num_days_no_floating_correction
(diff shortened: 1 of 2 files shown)


sfirke:
Looks good, thank @billdenney !

PR #223, opened 2018-08-02
remove_empty should work with more (and probably any) 2-dimensional classes

### LINKED ISSUE #215: Add remove_empty for matrices?
## Feature requests

I was surprised at the output of `remove_empty` when passed a matrix. It seems to return a vector rather than a trimmed matrix object:

```
mat <- matrix(c(NA, NA, NA, rep(0, 3)), ncol = 2, byrow = TRUE)

janitor::remove_empty(mat)

> [1] NA  0  0  0
```

Is adding functionality for matrices out of scope? If so, maybe it would be good to throw an error or warning if passed a matrix?

billdenney:
Fix #215

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,10 +1,16 @@
+# janitor 1.1.1.9000
+
+## Bug fixes
+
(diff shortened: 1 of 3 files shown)


sfirke:
This is a solid improvement, appreciate the tests of single-column inputs and results 👍

Issue #247, opened 2018-10-26
Feature Request: `clean_names()` method for simple features


JosiahParry:
`read_csv() %>% janitor::clean_names()` is almost always the first line of code that I run for any project. Lately I have been migrating my spatial analyses to R with the help of the `sf` package. Unfortunately simple features are not very receptive to the `clean_names()` function. 

I would like to request a `clean_names.sf` method. 

Running the follow line using [this repo](https://github.com/josiahparry/somerville_permits) produces an error.

```
st_read("data/Parcels/Parcels.gdb", layer = "Parcels") %>%
        janitor::clean_names()
```

```
Error in st_geometry.sf(x) : 
  attr(obj, "sf_column") does not point to a geometry column.
Did you rename it, without setting st_geometry(obj) <- "newname"?
```

billdenney:
@JosiahParry, that seems reasonable to me.

As I'm not familiar with `sf` objects, can you please specify how the behavior should differ?  From the error message, it suggests that we should do something like the pseudo code below (this is typed without actually looking at an `sf` object, and I rarely use NextMethod, so it probably doesn't work):

```r
clean_names.sf <- function(x, ...) {
  orig_names <- names(x)
  orig_sf_column <- st_geometry(x)
  ret <- NextMethod(clean_names)
  st_geometry(ret) <- names(ret)[orig_names %in% orig_sf_column]
  ret
}
```

... better yet, could you please provide a PR with a test?

JosiahParry:
I actually think I figured out a way to avoid using methods  (just because I'm naive and don't understand the intricacies quite yet).

This ought to work:

```
clean_names.sf <- function(x, ...) {
  ret <- clean_names(x)
  st_geometry(ret) <- names(ret)[ncol(ret)]
  ret
}
``` 

Will create a pr

sfirke:
Is there a way to do this that doesn't add the `sf` package as a dependency to janitor?  I'm not sure calling `sf::st_geometry()` without adding `sf` as a dependency will pass muster with CRAN.  Before I go asking around, anyone know?

sfirke:
(I like the idea of `clean_names` methods for common object types that are not `data.frame`)

sfirke:
Re: my comment about dependencies above, I got the advice:

> use `Suggests` field inside the `DESCRIPTION` file for the package. During dispatch to the `sf` object check if the namespace is present. e.g.
```
if(requireNamespace("sf", quietly = TRUE))  {
  # good
}
[...]

PR #248, opened 2018-10-26
Create sf method for `clean_names()`

JosiahParry:
Addressing issue https://github.com/sfirke/janitor/issues/247. 

This pull request creates a method of `clean_names()` to be used on `sf` objects.


sfirke:
Thanks for this!  I have never used `sf` objects myself but know they are widely-used and appreciate this PR.  I'd like to get this merged in.

The development version of janitor on GitHub now has a function `make_clean_names` that takes a character vector of names and returns a character vector of clean names.  That seems useful for a function like this, since `clean_names` is meant for a data.frame and `make_clean_names` is leaner.  Could `make_clean_names()` simplify this?

Maybe something like:
```
st_geometry(x) <- dplyr::last(make_clean_names(names(x)))
x
```
If `sf` objects have `names` that is.  Might need to be edited, I have not tried it.

PR #261, opened 2019-01-01
Add chisq.test function for two-way tabyl objects

### LINKED ISSUE #255: Add chi square test for tabyl objects?
To offer all the functionality of table(), would it be useful to be able to run a chisq.test() on a tabyl?  Would other statistical tests be useful?

Might be almost this simple, no?
```
chisq.test.tabyl <- function(dat, ...){
  # should check if input is two-way tabyl
  dat %>%
    dplyr::select(-1) %>%
    as.matrix() %>%
    chisq.test(...)
}
```

juba:
## Description
This PR aims to apply the chisq.test function to two-way tabyls. It overrides the default stats::chisq.test function, tests if the passed table is a two way tabyl and then calls a new chisq.test.tabyl function. Otherwise it just normally calls stats::chisq.test.

By default, chisq.test.tabyl also converts the tables in the returned object (`observed`, `expected`, `residuals`, `stdres`) as tabyls. This can be disabled with the `tabyl_results` argument.

Tests have been included.

## Related Issue
Fixes #255

## Example
tab <- tabyl(mtcars, gear, cyl)
chisq.test(tab)
chisq.test(tab)$residuals
chisq.test(tab, tabyl_results = FALSE)$residuals


sfirke:
Sorry again for the delay in reviewing!  My big Q is around overwriting the base function `chisq.test`.  Does that accomplish something that is missing if we just have `chisq.test.tabyl`?   You have `chisq.test` checking to see if it should call the `.tabyl` version but I think the point of S3 is to accomplish that automatically based on object type.

I looked on CRAN to see if there were examples of other authors writing S3 methods for this function.  This might be a model we want to use? https://github.com/chgigot/epiphy/blob/97ebd950e0a6471f031e7c7c84330776e4be6b57/R/indices.R#L329

sfirke:
Oh and smaller things but could you update the NEWS.md file to note your contribution, and add this to an appropriate place in the tabyls vignette?  Noting that the chisq.test function can be called on `tabyl` data.frames, resulting in `tabyl` outputs.  Thanks again!

juba:
Hi,

I think there is no choice to overwrite the default `stats::chisq.test` function, because it is not an S3 function. In the example you give in the `epiphy` package, they do rewrite `chisq.test` too, because otherwise there would be no way to make `chisq.test` apply `chisq.test.tabyl` on `tabyl`-class objects.

The difference is that in `epiphy` they indeed rewrite `chisq.test` as an S3 function, by creating `chisq.test.default`, and use `useMethod` in the rewritten `chisq.test`. If you think it would be better, I can rewrite my PR in the same way.

I'll also try to update my PR for NEWS.md and the vignette.

Thanks for your feedback !

sfirke:
I appreciate your clarifying restatement, that is spot on.  Yes, I think re-writing `chisq.test` as an S3 method as in the `epiphy` package is preferable - but I don't feel very confident in my opinion.  @chrishaid you were the one who taught me about S3, do you have thoughts here on the best way to implement `chisq.test.tabyl`?

sfirke:
@billdenney do you have thoughts on the better way to implement this S3 method, so @juba can proceed? I did not hear from Chris.  It would be great to add this in the next week so it is included in v1.2 going to CRAN by 3/20.

juba:
After thinking about it a little, I think it would be better to reimplement this as S3 methods. As such I think it could be compatible with potential other implementations such as the one in epiphy.

sfirke:
Agreed, good point. Can you take a stab at reworking it that way this week?

juba:
I just rewritten `chisq.test` as an S3 method, and I did the same for a new `fisher.test` function, which also applies to two-way tables.

As you requested, I also (very briefly) updated `NEWS.md` and the `tabyls`  vignette to mention these two new functions.

I hope this PR is ok, don't hesitate to tell me if something is not right.

billdenney:
Hey @juba, thanks for the PR.  I didn't have a chance to review everything in detail, but I do have a suggestion for making these methods into S3 methods.

For the S3 default methods, we would want to point them back to the standard `stats::` methods so that someone who loads `janitor` would still get the expected results for anything other than a tabyl.  To do that, please update the PR to do the exact same thing that epiphy does:

```r
#' @rdname chisq.test
#' @method chisq.test default
#' @export
#' @importFrom stats chisq.test
chisq.test.default <- function(x, ...) stats::chisq.test(x, ...)
```

You'd want to do the same thing with `fisher.test.default()` (just substituting "fisher" for "chisq" in the above):

```r
#' @rdname fisher.test
#' @method fisher.test default
#' @export
#' @importFrom stats fisher.test
fisher.test.default <- function(x, ...) stats::fisher.test(x, ...)
```

juba:
Hi, and thanks for your feedback.

Well, I think that's what my PR is doing. There's just one supplementary operation which consists of getting the name of the passed table object with `deparse(subtitute())` in order to have the same `data.name` value as in the `stats::chisq.test` or `stats::fisher.test` results.

billdenney:
@juba, I just looked in more detail, and the underlying issue I was trying to describe is that if someone calls `chisq.test()` in any scenario other than for a `tabyl`, it should return the same thing.  For that, using the single-line default method will automatically do it.  The fact that the second call to `identical()` below is `FALSE` will be an issue for some people; we should not change the behavior of a default call:

``` r
chisq.test.default <- function(dat, tabyl_results = TRUE, ...) {
  
  # keep track of object name to keep `data.name` attribute
  dname <- deparse(substitute(dat))
  
  result <- stats::chisq.test(dat, ...)
  result$data.name <- dname
  
  result
}

x <- matrix(c(12, 5, 7, 7), ncol = 2)
identical(stats::chisq.test(x), chisq.test.default(x))
#> [1] TRUE

x_vec <- rep(1:2, each=3)
y_vec <- 1:6
identical(stats::chisq.test(x_vec, y_vec), chisq.test.default(x_vec, y_vec))
#> Warning in stats::chisq.test(x_vec, y_vec): Chi-squared approximation may
#> be incorrect
#> Warning in stats::chisq.test(dat, ...): Chi-squared approximation may be
#> incorrect
#> [1] FALSE
```

Created on 2019-03-11 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

juba:
Ah, yes, I didn't think of the case when you pass two vectors instead of a two-way table. Once again, I can't use the single line way because in this case the two results will not be identical due to the `data.name` result field. But I'll take a look at the two vectors case and I'll add a test too.

Thanks for pointing this out !

juba:
So, I changed the PR i order to handle the case when two factors are passed to `chisq.test` or `fisher.test` as `x` and `y` arguments instead of a two-way table. The resulting code may seem overly complicated, but I think it is the only way to ensure a perfect equality between resulting object with `stats::chisq.test` and `stats::fisher.test`. For example, the approach taken by `epiphy` is much more simple, but the result objects are not exactly identical because of the `data.name` attribute.

``` r
library(epiphy)
#> 
#> Attachement du package : 'epiphy'
#> The following object is masked from 'package:stats':
#> 
#>     chisq.test
tab <- table(mtcars$cyl, mtcars$am)
res <- stats::chisq.test(tab)
#> Warning in stats::chisq.test(tab): Chi-squared approximation may be
#> incorrect
res2 <- epiphy::chisq.test(tab)
#> Warning in stats::chisq.test(x, ...): Chi-squared approximation may be
#> incorrect
identical(res, res2)
#> [1] FALSE
testthat::expect_equal(res, res2)
#> Error: `res` not equal to `res2`.
#> Component "data.name": 1 string mismatch
```

It is worst in the case of two vectors, as the dimnames of the `observed`, `expected`, etc. tables in the result object are not identical either :

``` r
res <- stats::chisq.test(mtcars$gear, mtcars$cyl)
#> Warning in stats::chisq.test(mtcars$gear, mtcars$cyl): Chi-squared
#> approximation may be incorrect
res2 <- epiphy::chisq.test(mtcars$gear, mtcars$cyl)
#> Warning in stats::chisq.test(x, ...): Chi-squared approximation may be
#> incorrect
identical(res, res2)
#> [1] FALSE
testthat::expect_equal(res, res2)
#> Error: `res` not equal to `res2`.
#> Component "data.name": 1 string mismatch
#> Component "observed": Attributes: < Component "dimnames": Names: 1 string mismatch >
#> Component "expected": Attributes: < Component "dimnames": Names: 1 string mismatch >
#> Component "residuals": Attributes: < Component "dimnames": Names: 1 string mismatch >
#> Component "stdres": Attributes: < Component "dimnames": Names: 1 string mismatch >
```

So if we want to keep a strict equality of results, I think we are forced to keep track of the names of the objects passed as arguments and to replace them before returning the result. The alternative would be to not care about this, resulting in a simpler code but result objects slightly different by names.

I hope all of this is clear enough... Thanks again for your feedback.

billdenney:
@juba, I appreciate the effort you're putting in for this!  Thank you!

One additional note, we would want to setup the S3 generic so that anyone creating a new generic off of our generic would have the logically-required generics (I may have been able to write a more confusing sentence, but I doubt it. :) )  Essentially, we should minimize the arguments to the generic function definition and only apply the arguments that would always be required.

So, could you please update the default arguments to be:

```r
chisq.test <- function(x, ...) {
  UseMethod("chisq.test")
}

chisq.test.default <- function(x, ...) {
  
  # keep track of object name to keep `data.name` attribute
  dname <- deparse(substitute(x))
  
  result <- stats::chisq.test(x=x, ...)
  result$data.name <- dname
  
  result
}
```

(With the same modifications done to `fisher.test()`.)

That way, if someone specifies the argument name when calling `chisq.test(x=my_data)`, it will still work.  And, if someone else wants to make a `chisq.test()` function for their method, they are only required to handle an `x` argument (not the `tabyl_results` which would only apply for `tabyl`s).

I hope that these comments don't come across as too nit-picky.  The hope is that we will work nicely with existing code.  If we break existing code, chances are good that we'd need to change it in the future.

juba:
Ah, yes, you're right, that's a perfectly good point. Nothing nit-picky here, I really appreciate your useful feedback. I'll take a look at it tomorrow or wednesday.

Thanks !

sfirke:
Thanks @juba for keeping this going and being receptive to feedback, and
@billdenney I appreciate you illustrating your point with a suggested code
sample.  I am fairly inexperienced with S3 methods myself honestly and I am
grateful to you two for going a few rounds on this to truly get it right.
I think this is a shining example of open collaboration!

Bill, you should feel free to merge this when you think it's ready, no need
for me to weigh in.

juba:
So here's a new commit that move specific arguments `y` and `tabyl_results` to their methods (respectively `chisq.test.default` and `chisq.test.tabyl`) instead of the generic S3 function.

So here are the new function signatures : 

```r
chisq.test <- function(x, ...)
chisq.test.default <- function(x, y = NULL, ...)
chisq.test.tabyl <- function(x, tabyl_results = TRUE, ...)

fisher.test <- function(x, ...)
fisher.test.default <- function(x, y = NULL, ...)
fisher.test.tabyl <- function(x, ...)
```

I don't think I can remove `y` from the `.default` methods, otherwise I can't find a way to reliably test the existence and get the value of it. I don't think this would be a problem for another `chisq.test()` method from another package anyway.

I hope the PR will be good enough now, but I'm obviously open to any new feedback.

Thanks a lot again.

billdenney:
I think that looks right.  Thanks for your persistence!

PR #272, opened 2019-02-19
Remove constant

### LINKED ISSUE #222: Feature Request: Drop Single-Value Columns
I often receive data where many of the columns may be a single value.  It's helpful find the information of interest by dropping all columns that only have a single value.  Would you be interested to include this feature in `janitor`?

``` r
interesting_cols <- function(x, ...) {
  mask_boring <-
    sapply(x,
           FUN=function(y) all(y %in% y[[1]]))
  boring <- x[1,mask_boring]
  ret <- x[,!mask_boring]
  attr(ret, "boring") <- boring
  ret
}
```

billdenney:
Fix #222


billdenney:
@sfirke, I accidentally pushed this to master, sorry!

While it was an accident, I think that we were aligned on the implementation, so I'm not immediately rolling it back.  Please roll back if you have any concerns.

PR #281, opened 2019-03-15
Update timezone docs (Fix #233)

### LINKED ISSUE #233: Add further reading references to 'tz' argument of excel_numeric_to_date
Related to #232, I think we should provide a reference link in the documentation for the `tz` argument that would allow users to read more about valid timezones and see a list of them.  I found my way to `https://data.iana.org/time-zones/tz-link.html` but would like something simpler. A list of common timezones on a website might be more useful to the typical user than shell commands to download a tarball.

Maybe two links, the one I have above for further / more advanced reading and a simpler one with list of common timezones?

billdenney:
Fix #233

I did text wrapping which makes this look like a much larger diff.  The only real change is:

```r
#' A list of all timezones is available from \code{base::OlsonNames()}, and the
#' current timezone is available from \code{base::Sys.timezone()}.
```

### DIFF
--- a/R/excel_dates.R
+++ b/R/excel_dates.R
@@ -1,21 +1,35 @@
 #' @title Convert dates encoded as serial numbers to Date class.
 #'
-#' @description
-#' Converts numbers like \code{42370} into date values like \code{2016-01-01}.
+#' @description Converts numbers like \code{42370} into date values like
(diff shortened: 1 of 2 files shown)


sfirke:
👏

PR #303, opened 2019-05-20
improve readability of make_clean_names

Tazinho:
I'd like to make some small adjustments regarding the readability of the `make_clean_names()` function. I think these might make it more easy to reason about current and further upcoming issues. The adjustments include:
* Some smaller adjustments in the beginning (explicit comments and removing of the pipe)
* Including also unused arguments of `to_any_case()` explicitly in the source code. This might make it easier to reason about issues, which might get solved by providing snakecase arguments to janitor users.

### DIFF
--- a/R/make_clean_names.R
+++ b/R/make_clean_names.R
@@ -34,49 +34,55 @@
 make_clean_names <- function(string, case = c(
   "snake", "lower_camel", "upper_camel", "screaming_snake",
   "lower_upper", "upper_lower", "all_caps", "small_camel",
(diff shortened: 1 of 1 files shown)


sfirke:
@Tazinho: the overhaul of `clean_names` happening in #340 makes me think we should close this pull request.  I think you'll like Bill's changes, among other things all of the arguments to `to_any_case` are exposed to `clean_names` via `...`.  And we'll add examples of using `abbreviation` and cases like Title Case.

Want to take a look at #340 as it stands and, if it looks like it covers what you have here, close this one?

PR #305, opened 2019-06-24
Remove empty quiet arg

### LINKED ISSUE #70: verbose options to remove_empty_ functions
Hi @sfirke ,

Thanks for this package. Works really well in first step of every analysis.

A lot of times you don't kown a priori your load data set but anyway you want to clean it with the `remove_empty_` functions ,and sometimes we want to know what/how much we removed. 

```
options(janitor.verbose = TRUE)

data <- data_raw %>%
  remove_empty_rows() %>%
  remove_empty_cols()

# as messages
3.234 (45%) rows removed
4 (5%) columns removed (colna1, colna2, ...)
```

Do you think this useful to add to your janitor package?

billdenney:
Add `quiet` argument to remove_* (Fix #70)

### DIFF
--- a/tests/testthat/test-remove-empties.R
+++ b/tests/testthat/test-remove-empties.R
@@ -130,6 +130,11 @@ test_that("Messages are accurate with remove_empty and remove_constant", {
     fixed=TRUE
   )
   expect_message(
+    remove_empty(data.frame(A=NA, B=1, C=NA), which="cols", quiet=FALSE),
+    regexp="Removing 2 empty columns of 3 columns total (Removed: A, C).",
+    fixed=TRUE
+  )
+  expect_message(
     remove_empty(data.frame(A=NA, B=c(1, NA)), which="rows", quiet=FALSE),
     regexp="Removing 1 empty rows of 2 rows total (50%).",
     fixed=TRUE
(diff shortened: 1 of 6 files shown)


tests/testthat/test-remove-empties.R
```diff
@@ -122,3 +122,27 @@ test_that("remove_constant", {
     info="NA with other values is kept with na.rm"
   )
 })
+
+test_that("Messages are accurate with remove_empty and remove_constant", {
+  expect_message(
+    remove_empty(data.frame(A=NA, B=1), which="cols", quiet=FALSE),
+    regexp="Removing 1 empty columns of 2 columns total (Removed: A).",
+    fixed=TRUE
```
sfirke: I could see adding one test here where multiple columns are removed, to test how the delimiting of those names works.

sfirke:
Looks good to me 👍  Thanks for implementing.

Appears the Travis build is failing b/c two chisq.test examples are throwing the warning `Chi-squared approximation may be incorrect` ? 

You can go ahead and merge this, either now or after adding one more petty test for a trivial little comment I raised (adding a test for the case where multiple columns are removed, I wanted to check how they would be listed, e.g., as bullets or commas).

I'm heading offline for a bit so you can feel free to merge or do as you see fit without any additional review from me!

billdenney:
Good catch, and I also added a couple to ensure that silence is maintained when `quiet=TRUE`.

PR #308, opened 2019-07-14
Simplify the logic within remove_constant

billdenney:
This doesn't change functionality, but it simplifies the code within `remove_constant()`.  I had an error in some of my code that I thought was related to an issue with `remove_constant()` (it wasn't remove_constant's fault), but then I tried to look at the code within `remove_constant()`, and I found it difficult to read-- so I simplified it.

Since this is not a functional change, I don't think it needs a NEWS entry, but let me know if you think otherwise.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,7 @@
 ## Bug fixes
 
 * The `name` argument to `adorn_totals()` is correctly applied to 3-way tabyls (#306)  Thanks to **@jzadra** for reporting.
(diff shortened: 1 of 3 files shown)


billdenney:
... it turns out there was a bug after all.  The issue is that matrices require indexing like `dat[, idx]` while tibbles require indexing like `dat[[idx]]`.

sfirke:
whoa that was actually a pretty big 🐛, given how much I presume most `janitor` users rely on tibbles. Nice catch & patch! ✨

PR #309, opened 2019-07-14
Correct warnings for deprecated tibble functions; skip tests when sf is not installed

billdenney:
When running the tests, I got warnings for `data_frame()` and `as_data_frame()` being deprecated, so I switched to the `tibble` equivalent.

Also, I do not have the `sf` package installed, so I get an error in testing the `sf` functions.  I think that I updated those tests correctly, but it is probably worth a check on a system with `sf` installed to ensure that there are no skipped tests and all tests succeed.

### DIFF
--- a/tests/testthat/test-add-totals.R
+++ b/tests/testthat/test-add-totals.R
@@ -74,7 +74,7 @@ test_that("order doesn't matter when row and col are called together", {
 })
 
 test_that("both functions work with a single column", {
-  single_col <- data_frame(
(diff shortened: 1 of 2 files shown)


sfirke:
Looks great!  I did not know about `skip_if_not_installed` but that seems perfect for this case.  And I see dplyr re-exports `tibble()` and `as_tibble()` so we're good there.

I'll merge it, then check on a machine later that has `sf` installed.

PR #311, opened 2019-07-25
Add convert_to_date (fix #310)

### LINKED ISSUE #310: Feature Request: Make excel_numeric_to_date() Generic
Currently, `excel_numeric_to_date()` requires numeric input, but as is the case with many `janitor` functions, input is often messy.

I often get files that have various formats for date and sometimes I want to use generic code to clean them up.  With that, I'd like to have `excel_numeric_to_date()` be able to take in POSIXt or Date classes.  The return value would be an appropriate of either POSIXct (if `include_time=TRUE`) or Date (if `include_time=FALSE`).

The downside to this is that things other than numbers are no longer is actually "excel numeric" to date, but a more general to date function.

If interested, I'll make a PR.

billdenney:
This implements the changes suggested in #310 handling numeric, character, factor, Date, and POSIXt classes for both Date and POSIXct output.


R/convert_to_date.R
```diff
@@ -0,0 +1,151 @@
...
+      )
+    }
+    ret[mask_character] <- characters_converted
+    if (any(is.na(ret[mask_character]))) {
+      not_converted_values <- unique(x[mask_character & is.na(ret)])
+      not_converted_message <-
+        paste0(
+          "Not all character strings converted to class ", out_class,
+          ".  Values not converted were: ",
+          paste0('"', not_converted_values, '"', collapse=", ")
```
sfirke: Is it possible/likely that this could be say, a thousand values, if this function were called incorrectly and would that cause a crash?  Maybe truncate at some point - there's code in `get_dupes` for that?

R/convert_to_date.R
```diff
@@ -0,0 +1,151 @@
...
+    out_class="Date"
+  )
+}
+
+#' @describeIn convert_to_date Convert to a date-time (POSIXct)
+#' @examples
+#' convert_to_datetime(c("2009-07-06", "40000.1", "40000", NA), character_fun=lubridate::ymd_h, truncated=1, tz="UTC")
+#' @export
+#' @importFrom lubridate ymd_hms
+convert_to_datetime <- function(x, ..., tz="UTC", character_fun=lubridate::ymd_hms, allow_na_conversion=FALSE) {
```
sfirke: `allow_na_conversion` seems to me to mean, allow NAs to be converted.  What do you think about instead, `string_conversion_failure = c("error", "warning")` or similar for that argument?  Then change @param documentation above to be something like, "if a character value fails to parse into the desired class and instead returns NA, should the function return the result with a warning or throw an error?"

R/excel_dates.R
```diff
@@ -38,6 +38,7 @@
 #' excel_numeric_to_date(40000.521, include_time = TRUE) # Time is included
 #' excel_numeric_to_date(40000.521, include_time = TRUE,
 #'   round_seconds = FALSE) # Time with fractional seconds is included
+#' @family Date conversion
```
sfirke: Let's put a longer note here like "if your input column(s) have a mix of Excel numeric dates and actual dates, see the more powerful functions `convert_to_date` and `convert_to_datetime`."

tests/testthat/test-convert_to_date.R
```diff
@@ -0,0 +1,127 @@
...
+    as.Date("2009-07-06")
+  )
+  expect_equal(
+    convert_to_date(c("2009-07-06", "40000.1", "40000", NA)),
+    c(rep(as.Date("2009-07-06"), 3), NA),
+    info="Mixed input works, including NA."
+  )
+})
+
+test_that("convert_date works", {
```
sfirke: should be "convert_datetime" works

sfirke:
I deeply apologize for letting this sit unreviewed for so long.  It's carefully-written, powerful but not heavy-handed.  I think it will be a valuable tool for data janitors 😁   Thanks for the contribution!  Let me know what you think of the feedback & if anything is not doable or is off-base.

sfirke:
Doh I re-clicked "review" but I'm done.  Also I see it now has a merge conflict with the master branch DESCRIPTION, that shouldn't be bad to resolve line by line but it's one more item I suppose.

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # janitor 1.2.0.9000 (unreleased)
 
+## Major features
+
+* The new functions `convert_to_date()` and `convert_to_datetime()` generalize the work done by `excel_numeric_to_date()` allowing conversion to date or datetimes from many forms of input from numeric, to characters that look like numbers, to characters that look like dates or datetimes, to Dates, to date-times (POSIXt) (#310, thanks to **@billdenney* for implementing).
```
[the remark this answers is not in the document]
billdenney: :)

R/convert_to_date.R
```diff
@@ -0,0 +1,151 @@
...
+    character_fun=character_fun,
+    allow_na_conversion=allow_na_conversion,
+    out_class="Date"
+  )
+}
+
+#' @describeIn convert_to_date Convert to a date-time (POSIXct)
+#' @examples
+#' convert_to_datetime(c("2009-07-06", "40000.1", "40000", NA), character_fun=lubridate::ymd_h, truncated=1, tz="UTC")
+#' @export
```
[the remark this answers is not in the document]
billdenney: I added one.

tests/testthat/test-convert_to_date.R
billdenney: Fixed

R/convert_to_date.R
billdenney: Good call.  I added that.

R/convert_to_date.R
billdenney: That is easier to understand.  Good suggestion!

R/excel_dates.R
billdenney: Good call.  Updated.

sfirke:
I think the branch build is failing due to this:
```
 checking dependencies in R code ... WARNING
'::' or ':::' import not declared from: ‘tidyselect’
```
Is there a function that needs the `tidyselect::` in front of it?
Might as well address that here, but once this passes on Travis-CI then it's :+1: :+1: from me and you're good to squash and merge!

billdenney:
I accidentally dropped `tidyselect` from the `DESCRIPTION` file.  When the Travis build completes confirming that I resolved it (`devtools::check()` looks good on my local system), then I'll merge.

PR #315, opened 2019-08-11
Improve compare_df_cols error

billdenney:
When running `compare_df_cols()` today on about 10 data.frames, one of them was the wrong class.  I didn't want to manually review all of them, so I made the attached update to point me to the one that was problematic.

All it does is gives a specific error message pointing the user to the argument numbers that are problems (and then tests that).


sfirke:
Sorry I missed this PR, the improve error catching for compare_df_cols.  This looks like 3 commits that handle two different issues?  I am ready to merge the error catching + messaging but should I review the date-handling aspect as well?

billdenney:
It was a mea-culpa to combine the two.  Let me regenerate this as a separate PR.  (I think that in the end, this PR will go away to be replaced.)

billdenney:
Please review #323

PR #321, opened 2019-12-13
Enable `row_to_names()` for matrix input

### LINKED ISSUE #320: Feature/Bug: row_to_names does not work on matrices
## Feature requests

`row_to_names()` should work on matrices.

## Bug reports

When `row_to_names()` is applied to a matrix, it silently drops the header row and rows above it without changing the name.

``` r
library(janitor)
#> 
#> Attaching package: 'janitor'
#> The following objects are masked from 'package:stats':
#> 
#>     chisq.test, fisher.test
row_to_names(
  matrix(LETTERS[1:4], nrow=2),
  row_number=1
)
#>      [,1] [,2]
#> [1,] "B"  "D"
```

<sup>Created on 2019-12-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

billdenney:
Fix #320

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,7 @@
 ## Minor features
 
 * A `quiet` argument was added to `remove_empty()` and `remove_constant()`  providing more information (when `FALSE`) (#70, thanks to **@jbkunst** for suggesting and **@billdenney** for implementing).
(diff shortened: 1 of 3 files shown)


sfirke:
Thanks for catching and fixing, I'll take a look during the week and merge!

sfirke:
Thanks Bill, looks great!

PR #323, opened 2019-12-16
Give more information when classes are incorrect for compare_df_cols

billdenney:
replaces #315


sfirke:
Nice, thanks for breaking this out!

Issue #314, opened 2019-08-09
feature request: signif_half_up()


khueyama:
I'm currently working on a project that needs halves away from zero rounding and was thankful to come across `janitor::round_half_up()`! Part of this project also requires similar behavior for significant digit rounding, e.g., `signif(12.5, 2)` returns 12 instead of 13. I have an implementation of  `signif_round_up()` working and can put in a PR if it would be of interest to you.

billdenney:
@sfirke should weigh in, but I would think it would fit with the theme of the package.

sfirke:
Sorry to be slow replying! Yeah that sounds in scope. If you have a PR with
tests and documentation that would be appreciated. Let us know if you have
questions or want input. Bill, could you review such a PR?

billdenney:
I'd be happy to review a PR for this.

khueyama:
Great, thanks @sfirke & @billdenney! Have one last push in the office to get through before the holidays but then things will be nice and quiet. Will get things wrapped up and put in the PR sometime next week.

sfirke:
You two are great!  To me, this issue's set of suggestion/pull request/code review/feedback/revision exemplifies the promise of open-source collaborative software.  This is a nice addition to the package and I'm grateful to both of you Kyle and Bill for adding it with such attention to detail in #324.

Happy New Year! :tada:

PR #324, opened 2019-12-27
implement signif_half_up functionality

khueyama:
## Description
@billdenney As discussed, this PR adds the `signif_half_up()` function. This function will round a numeric vector to the specified number of significant digits with halves rounded up instead of to even as in `signif()`. Let me know if any edits/additions are necessary. Thanks again!

## Related Issue
Implements #314.

## Example
`signif_half_up(12.5, digits = 2)` will return 13, whereas `signif(12.5, digits = 2)` will return 12. Negative numbers will round away from 0; `signif_half_up(-2.5, 1)` will return -3.

Tests implemented and currently passing at: tests/testthat/test-signif_half_up.R


billdenney:
Thanks for the PR!

For this type of function, issues often arise for the unusual numbers.  Can you please fix it so that it works for `NA`, `NaN`, `-Inf`, and `Inf`?  They should all return themselves unmodified to align with the standard base functions.  Please include an additional vector test that confirms that each of those works when in the presence of other numbers (like `c(1, 1.5, 1.49, NA, NaN, -Inf, Inf)` as the input).  For detection of infinite numbers, please use either `is.infinite()` or `is.finite()`.

Since the function is so closely related to `round_half_up()` please also move the code from a new .R file to `round_half_up.R` (that will make it easier to find in the future).

Finally, please also add it to the `NEWS.md` file giving yourself credit as you can see from the other notes there.

khueyama:
@billdenney Thank you for the feedback! I've updated as requested, let me know if you need anything else.

billdenney:
Looks good!  Thanks for the responsiveness.

I'm going to go ahead and merge this.  For future PRs, please always rerun `devtools::document()` to update the docs on your code (the only change is the file that the man page sources from).

Issue #326, opened 2020-01-16
Suggestion: allow get_dupes() to use tidyselect


jzadra:
`get_dupes()` allows you to specifiy columns positively, but you can't negate them (ie `get_dupes(mpg, -cty)` does not work, but `get_dupes(mpg, cty)` does.  

Often there's a single column that serves as an index or key that you know won't have a dupe, so it would be best to remove it rather than specifying all the vars except for it.

sfirke:
This is a no-brainer. Good improvement. Anyone is welcome to claim this
issue and submit a PR!

jzadra:
What the hell. I’ve never done one before and it is high time I learn. I think this is something I can do so I’ll give it a shot.

jzadra:
It's coming along.  First issue I've encountered that requires others input:

When using selection helpers like `starts_with()`, `ends_with()`, `contains()`, etc, there is the possibility that:

1. They catch nothing.  In this case I suggest we through the same error as you used before when there were no columns that matched the variables.
2. Some of the selections match columns, but others don't.  For instance consider that `mtcars %>% select(starts_with("nothere"), mpg)` will just return the mpg column.  Not a problem when you immediately see what you get.  But when expecting to get dupes, it might not be readily apparent that you didn't get anything matching "nothere" given that the return is the full data.frame plus the dupe_count column.  How do we want to handle this?  Should we just add a message indicating which columns were caught by tidyselect, or something else?

sfirke:
First a broad comment, I'm now looking at the code of `get_dupes` and I think you'll want to rewrite even this first bit to be tidyverse handling of `...` instead:
```
uq_names <- as.list(substitute(list(...)))[-1L]
    df_name <- deparse(substitute(dat))
```
Writing functions with tidyverse variable handling is not the easiest thing... but it's really useful to learn :stuck_out_tongue_winking_eye: 

My existing code is a hacky holdover from an earlier version of rlang.  I think on the upside that this stuff is better implemented and better documented now!

Re: your points, let's hold on the error + warning stuff for now.  I expect it might just be okay for say example #2 to use only `mpg` and the user can figure out what happened based on the fact that column 1 is `mpg` and column 2 is `n`.  If `dplyr::select` doesn't throw an error, I don't know that we need to.  We can always layer it in later but that's icing on the cake.

jzadra:
[...]

PR #328, opened 2020-01-17
Use tidyselect with get_dupes

jzadra:
Changed the method for selecting variables in the df to use `tidyselect`.  This allows negation, as well as `tidyselect helpers` like `starts_with()`, `ends_with()`, `contains()`, `matches()`, etc.

## Related Issue
#326 

## Example
`mtcars %>% get_dupes(weight = wt, starts_with("cy"))`

## Tests
Includes new tests:

```
test_that("tidyselect specification matches exact specification", {
  expect_equal(mtcars %>% get_dupes(contains("cy"), mpg), mtcars %>% get_dupes(cyl, mpg))
  expect_equal(mtcars %>% get_dupes(mpg), mtcars %>% get_dupes(-c(cyl, disp, hp, drat, wt, qsec, vs, am ,gear, carb)))
  expect_equal(suppressMessages(mtcars %>% select(cyl, wt) %>% get_dupes()), mtcars %>% select(cyl, wt) %>% get_dupes(everything()))
})
```

Also modifies test for specifying zero matching vars, as the old function for doing so was removed.  tidyselect now throws its own error if nothing matches for exact specification.  

NOTE however that using a tidyselect helper that doesn't catch anything will not produce any error or message.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -26,7 +26,8 @@ Imports:
     snakecase (>= 0.9.2),
     magrittr,
     purrr,
-    rlang
+    rlang,
+    tidyselect
 License: MIT + file LICENSE
 LazyData: true
 RoxygenNote: 7.0.2
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -26,8 +26,7 @@ Imports:
     snakecase (>= 0.9.2),
     magrittr,
     purrr,
-    rlang,
-    tidyselect
+    rlang
 License: MIT + file LICENSE
 LazyData: true
 RoxygenNote: 7.0.2
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -5,7 +5,8 @@ Authors@R: c(person("Sam", "Firke", email = "samuel.firke@gmail.com", role = c("
     person("Bill", "Denney", email = "wdenney@humanpredictions.com", role = "ctb"),
     person("Chris", "Haid", email = "chrishaid@gmail.com", role = "ctb"),
     person("Ryan", "Knight", email = "ryangknight@gmail.com", role = "ctb"),
-    person("Malte", "Grosser", email = "malte.grosser@gmail.com", role = "ctb"))
+    person("Malte", "Grosser", email = "malte.grosser@gmail.com", role = "ctb"),
+    person("Jonathan", "Zadra", email = "jonathan.zadra@sorensonimpact.com", role = "ctb"))
 Description: The main janitor functions can: perfectly format data.frame column
     names; provide quick counts of variable combinations (i.e., frequency
     tables and crosstabs); and isolate duplicate records. Other janitor functions
@@ -27,7 +28,7 @@ Imports:
     magrittr,
     purrr,
     rlang,
-    tidyselect
+    tidyselect (>= 1.0.0)
 License: MIT + file LICENSE
 LazyData: true
 RoxygenNote: 7.0.2
@@ -37,7 +38,5 @@ Suggests:
     rmarkdown,
     sf,
     tibble
-remotes:
-    r-lib/tidyselect
 VignetteBuilder: knitr
 Encoding: UTF-8
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -37,5 +37,7 @@ Suggests:
     rmarkdown,
     sf,
     tibble
+remotes:
+    r-lib/tidyselect
 VignetteBuilder: knitr
 Encoding: UTF-8
--- a/R/get_dupes.R
+++ b/R/get_dupes.R
@@ -3,47 +3,48 @@
 #' @description
 #' For hunting duplicate records during data cleaning.  Specify the data.frame and the variable combination to search for duplicates and get back the duplicated rows.
 #'
-#' @param dat the input data.frame.
-#' @param ... unquoted variable names to search for duplicates.
+#' @param dat The input data.frame.
+#' @param ... Unquoted variable names to search for duplicates. This takes a tidyselect specification.
 #' @return Returns a data.frame (actually a \code{tbl_df}) with the full records where the specified variables have duplicated values, as well as a variable \code{dupe_count} showing the number of rows sharing that combination of duplicated values.
 #' @export
 #' @examples
 #' get_dupes(mtcars, mpg, hp)
 #' # or called with the magrittr pipe %>% :
 #' mtcars %>% get_dupes(wt)
+#' # or using tidyselect:
+#' mtcars %>% get_dupes(weight = wt, starts_with("cy"))
 #'
 
 get_dupes <- function(dat, ...) {
-  uq_names <- as.list(substitute(list(...)))[-1L] # unquoted names for NSE calls, need quoted names separately for messages + warnings
-  df_name <- deparse(substitute(dat))
-
-  if (length(uq_names) == 0) { # if called on an entire data.frame with no specified variable names
+  
+  expr <- rlang::expr(c(...))
+  pos <- tidyselect::eval_select(expr, data = dat)
+  
+  names(dat)[pos] <- names(pos) #allows for renaming within get_dupes() consistent with select()
+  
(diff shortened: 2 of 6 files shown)


tests/testthat/test-get-dupes.R
```diff
@@ -37,3 +37,9 @@ test_that("works on variables with irregular names", {
   ) # does it return the right-sized result?
   expect_is(badname_df %>% get_dupes(), "data.frame") # test for success, i.e., produces a data.frame (with 0 rows)
 })
+
+test_that("tidyselect specification matches exact specification", {
+  expect_equal(mtcars %>% get_dupes(contains("cy"), mpg), mtcars %>% get_dupes(cyl, mpg))
+  expect_equal(mtcars %>% get_dupes(mpg), mtcars %>% get_dupes(-c(cyl, disp, hp, drat, wt, qsec, vs, am ,gear, carb)))
+  expect_equal(suppressMessages(mtcars %>% select(cyl, wt) %>% get_dupes()), mtcars %>% select(cyl, wt) %>% get_dupes(everything()))
```
billdenney: The `suppressMessages()` call suggests that there may be an issue here.  If a message or warning is desired, please use `expect_warning()` or `expect_message()` so that the test includes a test for expected behavior.  If a message or warning is not desired, please update the code so that the warning/message is not generated.

billdenney:
While I've not reviewed the code in detail, the continuous integration check with Travis CI indicates that there is an issue building the janitor.Rmd vignette (https://travis-ci.org/sfirke/janitor/jobs/638671020#L3940).  That suggests that there is some level of backward incompatibility in the change.

Can you please review what is occurring in the vignette and do the following:

* Add a test for the old functionality as occurred in the vignette
* Make sure that the new code works like the old code did (i.e. preserve backward compatibility)

billdenney:
I made a comment on your last test that I'm not sure if it will show up because there is a "pending" message above it.  If you get this twice, sorry, this is to ensure that you get it at least once:

The `suppressMessages()` call suggests that there may be an issue here.  If a message or warning is desired, please use `expect_warning()` or `expect_message()` so that the test includes a test for expected behavior.  If a message or warning is not desired, please update the code so that the warning/message is not generated.

tests/testthat/test-get-dupes.R
jzadra: The message from an empty `get_dupes()` call is a custom message that says `"No variable names specified - using all columns.\n"` (line 28).  This message doesn't occur when using `everything()` because you have specified variable names. My test was intended to compare the outputs of the two calls while ignoring that one had that custom message.  What's a better way to do this?

jzadra:
The only error I get when running the janitor.Rmd vignette is 

> While I've not reviewed the code in detail, the continuous integration check with Travis CI indicates that there is an issue building the janitor.Rmd vignette (https://travis-ci.org/sfirke/janitor/jobs/638671020#L3940). That suggests that there is some level of backward incompatibility in the change.
> 
> Can you please review what is occurring in the vignette and do the following:
> 
> * Add a test for the old functionality as occurred in the vignette
> * Make sure that the new code works like the old code did (i.e. preserve backward compatibility)

It looks like it was just that I forgot to add tidyselect to the imports in DESCRIPTION.  I've made the change on my fork, how do I add them to this PR?

billdenney:
There still appears to be an issue with the `janitor.Rmd` vignette.  Can you check that one?  Also, you can do most of the checks that will happen on Travis by running `devtools::check()`.

jzadra:
I've run devtools::check().  The only error I get is:

```
── 1. Failure: if called on non-existent vector, returns use
`mtcars %>% tabyl(moose)` threw an error with unexpected message.
Expected match: "object 'moose' not found"
Actual message: ""
```

sfirke:
Oh!  I think that error is caused by the newer tidyselect, this is why I have to resubmit to CRAN.  Let me see about patching that test right now.

sfirke:
I just committed this to master but try implementing it also in your branch: change line 190 of `test-tabyl.R` to be `  expect_error(mtcars %>% tabyl(moose))`

jzadra:
Yes that solved the moose error.  Interesting, I also had to change line 29 of test_get_dupes.R in the same way (removing the message).

jzadra:
It's still not finding tidyselect on travis.

sfirke:
The failing error message is:

> Quitting from lines 103-104 (janitor.Rmd) 
> 
> Error: processing vignette 'janitor.Rmd' failed with diagnostics:
> 
> 'eval_select' is not an exported object from 'namespace:tidyselect'
> 
> --- failed re-building ‘janitor.Rmd’

Looks like `eval_select` is only present in the dev version of tidyselect on GitHub?  There's a hit for it here: https://github.com/r-lib/tidyselect/search?q=eval_select&unscoped_q=eval_select and not here https://github.com/cran/tidyselect/search?q=eval_select&unscoped_q=eval_select (the current CRAN version).

Are you using a function that's only in the new version of tidyselect?  Ideally we wouldn't need to require the newest version of tidyselect, that would complicate getting both packages to CRAN.  If the same functionality could be implemented with existing tidyselect functions that aren't being made obsolete, that would be the best.

sfirke:
If you must use functions from upcoming tidyselect 1.0, or feel that they're superior and the way to go, it wouldn't be too bad to make janitor require tidyselect 1.0.  It doesn't have so many breaking changes that I think it would bother people to make them update it.

To do this with CRAN, I could submit a simple patch to handle the change in error message, then submit this feature as part of a larger 1.4 release following tidyselect 1.0 going to CRAN.

jzadra:
Looks like adding r-lib/tidyselect to the remotes has worked.  Would you prefer to wait until tidyselect 1.0.0 is on cran and set that version dependency in Imports?

tests/testthat/test-get-dupes.R
sfirke: You're good here then, thanks for explaining.

DESCRIPTION
```diff
@@ -26,7 +26,8 @@ Imports:
     snakecase (>= 0.9.2),
     magrittr,
     purrr,
-    rlang
+    rlang,
+    tidyselect
```
sfirke: let's specify at least version 1.0.0 now that it's on CRAN

DESCRIPTION
```diff
@@ -36,5 +37,7 @@ Suggests:
     rmarkdown,
     sf,
     tibble
+remotes:
+    r-lib/tidyselect
 VignetteBuilder: knitr
```
sfirke: and I think we can remove this now that 1.0.0 is on CRAN

R/get_dupes.R
```diff
@@ -3,47 +3,48 @@
...
   } else {
-    nms <- rlang::quos(...) %>%
-      rlang::quos_auto_name()
-    var_names <- names(nms)
+    var_names <- names(pos)
+    nms <- rlang::syms(var_names)
   }
-
-  # check that each variable name provided is present in names(dat); if not, throw error
-  check_vars_in_df(dat, df_name, var_names)
```
sfirke: is this removed b/c it was used to deal with how the old tidyverse functions errored and now they error in the desired way already -- i.e., they identify the mis-specified variable -- so it's now redundant?

jzadra:
Hi, sorry for the delay. I'm not sure how to integrate your changes with my local fork that is linked to the PR.  Is there a way to pull the changes you've made back into my local fork, or do I need to just manually copy paste from here into my fork? Some other method?  Sorry, first time!

R/get_dupes.R
jzadra: Correct.  Tidyselect will throw its own error if an exact column specification doesn't match the table.

But just a reminder (we'e discussed previously) that any matching functions (starts_with, etc) that don't catch any variables will not generate any error currently.

DESCRIPTION
jzadra: done.  And thanks for the invite to add myself as contrib!

DESCRIPTION
jzadra: done

jzadra:
Awesome, thanks for the coaching through my first PR and contribution!

PR #338, opened 2020-03-04
Add case argument to clean_names.sf

ari-nz:
## Description
Pass 'case' argument through to clean_names sf method.
Add test cases based on the regular data.frame method

## Related Issue
Issue at: #339
Related (extends) to #247 

## Example

This used to ignore the 'all_caps' argument
``` r
  library(janitor)
test_df = data.frame('column' = 'data')
test_df["long"] <- -80
test_df["lat"] <- 40

test_df <- sf::st_as_sf(test_df, coords = c("long", "lat"))

clean_names(test_df, 'all_caps')
#> Simple feature collection with 1 feature and 1 field
#> geometry type:  POINT
#> dimension:      XY
#> bbox:           xmin: -80 ymin: 40 xmax: -80 ymax: 40
#> epsg (SRID):    NA
#> proj4string:    NA
#>   COLUMN       geometry
#> 1   data POINT (-80 40)
```

### DIFF
--- a/R/clean_names.R
+++ b/R/clean_names.R
@@ -78,7 +78,7 @@ clean_names.sf <- function(dat, case = c(
   # identify ending column index to clean
   n_cols <- length(dat)-1 
   # clean all but last column
-  sf_cleaned <- make_clean_names(sf_names[1:n_cols]) 
+  sf_cleaned <- make_clean_names(sf_names[1:n_cols], case) 
   # rename original df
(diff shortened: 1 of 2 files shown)


sfirke:
This is great, thank you!  I'll approve + add this fix to the NEWS file, crediting you.  Also my son's name is Ari, always nice to run into another Ari in the world 😁

ari-nz:
No problem, a nice simple fix! Thanks for your work on the package so far!

Issue #251, opened 2018-11-16
clean_names() creates duplicate names


jzadra:
It seems that `clean_names()` can potentially create duplicate names as a product of cleaning them.  Running clean names a second time fixes it.  It might be good to check that there are no duplicates as a result of the cleaning, and if there are, re-run the process behind the scenes.

reprex:

``` r
require(tidyverse)
#> Loading required package: tidyverse
require(janitor)
#> Loading required package: janitor

w <- structure(list(unitid = "100690", f2a01 = "374175", f2a02 = "3123426", 
                    f2a03 = "1438541", f2a04 = "1522598", f2a05 = "162287", f2a06 = "1684885", 
                    f2a11 = "279448", f2a12 = "1028166", f2a13 = "1837548", f2a14 = "0", 
                    f2b01 = "2898929", f2b02 = "2814839", f2b03 = "0", f2b04 = "84090", 
                    f2b05 = "1600795", f2b06 = "0", f2b07 = "1684885", f2c01 = "223324", 
                    f2c02 = "11400", f2c03 = "0", f2c04 = "0", f2c05 = "0", f2c06 = "921452", 
                    f2c07 = "1156176", f2c08 = "921452", f2c09 = "0", f2d01 = "1987711", 
                    f2d02 = "0", f2d03 = "0", f2d04 = "0", f2d05 = "28770", f2d06 = "0", 
                    f2d07 = "0", f2d08 = "836984", f2d09 = "0", f2d10 = "28492", 
                    f2d11 = "0", f2d12 = "0", f2d13 = "0", f2d14 = "0", f2d15 = "16972", 
                    f2d16 = "2898929", f2e011 = "1242869", f2e012 = "696353", 
                    f2e021 = "0", f2e022 = "0", f2e031 = "0", f2e032 = "0", f2e041 = "84032", 
                    f2e042 = "58893", f2e051 = "279262", f2e052 = "175677", f2e061 = "1208676", 
                    f2e062 = "392969", f2e071 = "0", f2e072 = "0", f2e081 = "0", 
                    f2e082 = "0", f2e091 = "0", f2e092 = "0", f2e101 = "0", f2e102 = "0", 
                    f2e111 = "0", f2e112 = "0", f2e121 = "2814839", f2e122 = "1323892", 
                    f2e141 = "269211", f2e151 = "131249", f2e161 = "79703", f2e171 = "1010784", 
                    year = "1", f2e013 = NA_character_, f2e014 = NA_character_, 
[...]

PR #254, opened 2018-11-25
tbl_graph method for clean_names

JosiahParry:
Addressing issue https://github.com/sfirke/janitor/issues/252. 
Creating a method for `tbl_graph` objects.

PR #340, opened 2020-03-04
clean_names rewrite

### LINKED ISSUE #204: clean_names doesn't affect the delta symbol on OS X
On Windows clean_names() replaces the 'δ' symbol with 'd' but on OS X it is unaffected.

```r
library(janitor)

dat <- structure(list(`Sample ID` = "sample 1", `δ2H, in ‰` = -24.99), row.names = c(NA, 
                                                                                     -1L), class = c("tbl_df", "tbl", "data.frame"))
clean_names(dat)
```

### LINKED ISSUE #252: compatibility with tidygraph
I'm trying to explore the behaviour of `janitor` functions such as `clean_names()` on `tidygraph` objects as `tbl_graph`s. (I wrote about this previously [here](https://github.com/thomasp85/tidygraph/issues/71)

For an example, let's create a tiny test graph with bad named variables:

``` r
library(tidygraph)
#> 
#> Attaching package: 'tidygraph'
#> The following object is masked from 'package:stats':
#> 
#>     filter
library(janitor)
test_graph <- play_erdos_renyi(10, 0.5) %>% 
  mutate(Strange_Name  = 1,
         `Other bad Name` = 2)
```
The following cleaning silently doesn't work (columns names are left unchanged):

``` r
test_graph %>%
  janitor::clean_names()
#> # A tbl_graph: 10 nodes and 50 edges
#> #
#> # A directed simple graph with 1 component
#> #
#> # Node Data: 10 x 2 (active)
#>   Strange_Name `Other bad Name`
#>          <dbl>            <dbl>
#> 1            1                2
#> 2            1                2
#> 3            1                2
#> 4            1                2
#> 5            1                2
#> 6            1                2
#> # ... with 4 more rows
#> #
#> # Edge Data: 50 x 2
#>    from    to
#>   <int> <int>
#> 1     4     1
#> 2     6     1
#> 3    10     1
#> # ... with 47 more rows
```

On the other hand, and without much surprise, `clean_names()` works on the exported tibble:

``` r
test_graph %>%
  as_tibble() %>%
  janitor::clean_names()
#> # A tibble: 10 x 2
#>    strange_name other_bad_name
#>           <dbl>          <dbl>
#>  1            1              2
#>  2            1              2
#>  3            1              2
#>  4            1              2
#>  5            1              2
#>  6            1              2
#>  7            1              2
#>  8            1              2
#>  9            1              2
#> 10            1              2
```

It is not super clear to me why that happens. A simple `rename()` function as exported from `tidygraph` does the job, but that requires manual intervention:

``` r
test_graph %>%
  rename(lower_name = "Strange_Name",
         better_name = "Other bad Name")
#> # A tbl_graph: 10 nodes and 47 edges
#> #
#> # A directed simple graph with 1 component
#> #
#> # Node Data: 10 x 2 (active)
#>   lower_name better_name
#>        <dbl>       <dbl>
#> 1          1           2
#> 2          1           2
#> 3          1           2
#> 4          1           2
#> 5          1           2
#> 6          1           2
#> # ... with 4 more rows
#> #
#> # Edge Data: 47 x 2
#>    from    to
#>   <int> <int>
#> 1     1    10
#> 2     4     1
#> 3     5     1
#> # ... with 44 more rows
```
and that is the case even with a direct call to `dplyr::rename()`.

I am clearly missing something here. The same holds for the other `janitor` functions I tried.

Examples created on 2018-08-28 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

### LINKED ISSUE #268: clean_names inserts unwanted underscore "_" in between NON-ASCII characters
### Version

R version 3.5.1 (2018-07-02)
Platform: x86_64-apple-darwin15.6.0 (64-bit)

janitor 1.1.1

### Description

When you have  non-ascii column names, `clean_name`  inserts unwanted underscore in between NON-ASCII characters.

Below explains how unwanted underscore is inserted in between NON-ASCII characters.

``` r
df <- data.frame(`介護_看護_女` =  c(1))
colnames(df)
#> [1] "介護_看護_女"
df_cleaned <- janitor::clean_names(df)
colnames(df_cleaned)
#> [1] "介_護_看_護_女"
```

<sup>Created on 2019-02-08 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

For this example, clean_names shoule not put "_" in between characters like "介_護_看_護_女" but should return "介護_看護_女".

### LINKED ISSUE #271: clean names to proper names?
I've started working a little more with `kable` and rmarkdown. It would be really nice to have a quick parser that would take snake case names and make them into proper case names that would be suitable for column headers in a table. That way, if you operate within a snake case paradigm, it can take nice names for code and make them into nice names for tables. 

Imaginary example, assuming the function were called printable names:
```
iris %>% 
  clean_names() %>%
  proper_names() %>%
  names()
```
should output
```
"Sepal Length" "Sepal Width" "Petal Length" "Petal Width" "Species"
```
Or maybe the request is to have proper case added to [snake case](https://github.com/Tazinho/snakecase)? You could skip the `clean_names()` in the example if `proper_names()` wrapped a `snakecase` function. 

Tableau's parser does this automatically. It's really nice.

### LINKED ISSUE #283: clean_names: support abbreviations argument of `snakecase::to_any_case`?
Could `clean_names` support passing an `abbreviations` argument to the `abbreviations` argument of `snakecase::to_any_case`?

My use case for this would be to keep certain abbreviations e.g. "ID" capitalised when using `clean_names` to convert column names to UpperCamel.

``` r
library(magrittr)

# ID is converted to Id
tibble::tibble("person ID" = c(1)) %>%
  janitor::clean_names("upper_camel")
#> # A tibble: 1 x 1
#>   PersonId
#>      <dbl>
#> 1        1

# that's what happens by default using to_any_case
snakecase::to_any_case("Person ID", "upper_camel")
#> [1] "PersonId"

# but you can pass the abbreviations argument
snakecase::to_any_case("Person ID", "upper_camel", abbreviations = c("ID"))
#> [1] "PersonID"
```

<sup>Created on 2019-04-02 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>
<details><summary>Session info</summary>

``` r
devtools::session_info()
#> ─ Session info ──────────────────────────────────────────────────────────
#>  setting  value                       
#>  version  R version 3.5.2 (2018-12-20)
#>  os       macOS Sierra 10.12.6        
#>  system   x86_64, darwin15.6.0        
#>  ui       X11                         
#>  language (EN)                        
#>  collate  en_GB.UTF-8                 
#>  ctype    en_GB.UTF-8                 
#>  tz       Europe/London               
#>  date     2019-04-02                  
#> 
#> ─ Packages ──────────────────────────────────────────────────────────────
#>  package     * version date       lib source        
#>  assertthat    0.2.1   2019-03-21 [1] CRAN (R 3.5.2)
#>  backports     1.1.3   2018-12-14 [1] CRAN (R 3.5.0)
#>  callr         3.1.1   2018-12-21 [1] CRAN (R 3.5.0)
#>  cli           1.1.0   2019-03-19 [1] CRAN (R 3.5.2)
#>  crayon        1.3.4   2017-09-16 [1] CRAN (R 3.5.0)
#>  desc          1.2.0   2018-05-01 [1] CRAN (R 3.5.0)
#>  devtools      2.0.1   2018-10-26 [1] CRAN (R 3.5.1)
#>  digest        0.6.18  2018-10-10 [1] CRAN (R 3.5.0)
#>  evaluate      0.13    2019-02-12 [1] CRAN (R 3.5.2)
#>  fansi         0.4.0   2018-10-05 [1] CRAN (R 3.5.0)
#>  fs            1.2.6   2018-08-23 [1] CRAN (R 3.5.0)
#>  glue          1.3.1   2019-03-12 [1] CRAN (R 3.5.2)
#>  highr         0.7     2018-06-09 [1] CRAN (R 3.5.0)
#>  htmltools     0.3.6   2017-04-28 [1] CRAN (R 3.5.0)
#>  janitor       1.1.1   2018-07-31 [1] CRAN (R 3.5.0)
#>  knitr         1.21    2018-12-10 [1] CRAN (R 3.5.1)
#>  magrittr    * 1.5     2014-11-22 [1] CRAN (R 3.5.0)
#>  memoise       1.1.0   2017-04-21 [1] CRAN (R 3.5.0)
#>  pillar        1.3.1   2018-12-15 [1] CRAN (R 3.5.0)
#>  pkgbuild      1.0.2   2018-10-16 [1] CRAN (R 3.5.0)
#>  pkgconfig     2.0.2   2018-08-16 [1] CRAN (R 3.5.0)
#>  pkgload       1.0.2   2018-10-29 [1] CRAN (R 3.5.0)
#>  prettyunits   1.0.2   2015-07-13 [1] CRAN (R 3.5.0)
#>  processx      3.2.1   2018-12-05 [1] CRAN (R 3.5.0)
#>  ps            1.3.0   2018-12-21 [1] CRAN (R 3.5.0)
#>  R6            2.4.0   2019-02-14 [1] CRAN (R 3.5.2)
#>  Rcpp          1.0.1   2019-03-17 [1] CRAN (R 3.5.2)
#>  remotes       2.0.2   2018-10-30 [1] CRAN (R 3.5.0)
#>  rlang         0.3.1   2019-01-08 [1] CRAN (R 3.5.2)
#>  rmarkdown     1.11    2018-12-08 [1] CRAN (R 3.5.0)
#>  rprojroot     1.3-2   2018-01-03 [1] CRAN (R 3.5.0)
#>  sessioninfo   1.1.1   2018-11-05 [1] CRAN (R 3.5.0)
#>  snakecase     0.9.2   2018-08-14 [1] CRAN (R 3.5.0)
#>  stringi       1.4.3   2019-03-12 [1] CRAN (R 3.5.2)
#>  stringr       1.4.0   2019-02-10 [1] CRAN (R 3.5.2)
#>  testthat      2.0.1   2018-10-13 [1] CRAN (R 3.5.0)
#>  tibble        2.1.1   2019-03-16 [1] CRAN (R 3.5.2)
#>  usethis       1.4.0   2018-08-14 [1] CRAN (R 3.5.0)
#>  utf8          1.1.4   2018-05-24 [1] CRAN (R 3.5.0)
#>  withr         2.1.2   2018-03-15 [1] CRAN (R 3.5.0)
#>  xfun          0.5     2019-02-20 [1] CRAN (R 3.5.2)
#>  yaml          2.2.0   2018-07-25 [1] CRAN (R 3.5.0)
#> 
#> [1] /Library/Frameworks/R.framework/Versions/3.5/Resources/library
```

</details>

### LINKED ISSUE #316: feature request: custom translation of symbols such as % # in clean_names()
Currently clean_names() converts `%` to `percent` and `#` to `number`. Would it be possible to have an option like translate = c("%" = "pc", "#" = "no", "/" = "per")?

### LINKED ISSUE #331: `make_clean_names()` Drops Unicode First Characters
As the title suggests, `make_clean_names()` drops Unicode first characters.

```r
library(janitor)
make_clean_names(c("not_first_µg_dna", "µg_dna"))
# [1] "not_first_µg_dna" "g_dna"      
```

### LINKED ISSUE #339: clean_names() fails to change the case of columns of an sf data frame
## Feature requests

Enable clean_names to change the case of sf data frames

Related to #247

## Bug reports

`clean_names` doesn't respect the case argument in this case. the `column` should be `COLUMN`

```r
  library(janitor)
test_df = data.frame('COLUMN' = 'data')
test_df["long"] <- -80
test_df["lat"] <- 40

test_df <- sf::st_as_sf(test_df, coords = c("long", "lat"))

clean_names(test_df, 'all_caps')
#> Simple feature collection with 1 feature and 1 field
#> geometry type:  POINT
#> dimension:      XY
#> bbox:           xmin: -80 ymin: 40 xmax: -80 ymax: 40
#> epsg (SRID):    NA
#> proj4string:    NA
#>   column       geometry
#> 1   data POINT (-80 40)
```

## PR 

I've made a PR to solve this at #338

billdenney:
I'm vying to win the award for most issues closed with a single PR!  :)

Fixes #268
Fixes #271 
Fixes #283 
Fixes #331 
Fixes #339
Fixes #316 
Fixes #204
Fixes #252
Replaces #338
Replaces #254
Replaces #303

This rewrites `make_clean_names()` and `clean_names()` to address many issues, with the general goal of increasing flexibility and decreasing locale-dependence.

## Description

* All arguments to `make_clean_names()` are now accessible from `clean_names()`
* The `tbl_graph` method is now included in `clean_names()`
* General replacements are available for any text rather than just having the fixed replacements for `"'%#`
* `stringi` and `stringr` were used with Unicode-aware regular expression matching to be less locale-dependent.
* Hopefully, the only locale-dependent part is the use of `make.names()`, and that could possibly become optional or be removed to remove locale-dependence entirely.  (Removing it would remove functionality, so if removed, it would need to be replaced with a locale-independent version.)


sfirke:
Awesome, I will review as soon as I can.  If `make.names()` is locale dependent then I'd love to achieve the same thing in a more locale-independent way.  I appreciate you calling out guiding principles here, that change would feel in line with those too.

I just merged #338 (a) because they submitted it already before we got to this point (b) to welcome a new contributor.  But it looks like that causes a conflict with your pull request, sorry to make more work for you Bill.

Tazinho:
@billdenney @sfirke Hi guys, I believe the body of `make_clean_names()` can be even shorter (if you want to break some backword compatibility) and it could probably close some more issues. As I was not following the whole discussion regarding the rewrite of `make_clean_names` this solution might need to be adopted at some smaller prats like the `"Latin-ASCII"`as default.

Some of the advantages of the following would be
* no `make.names()` is involved (though one might need to test this for the special character behaviour as well before publishing this. Note I was not reading the whole discussion about the ASCII translation and reasons for this, but understood that it's somehow related to the locale and snakecase uses stringr internally and therefore typically only english locale)
* the boilerplate code for the translation is included in the snakecase call and the transliterations argument would fix #316 
* the unique_sep would fix #251 and also reduce the boilerplate code for adding the suffixes. (Note that this is a small breaking change as the counter for duplicates is different in `make.unique` which snakecase uses compared to the current implementation in janitor).

```r
library(snakecase)

string <- c("", "", "", 
            "bla #", "bla %", "so'so",
            "tatsächlich\"Liebe")

make_clean_names <- function(string, case = "snake", transliterations = NULL) {
  
  if(is.null(transliterations)) {
    transliterations <- c("#" = "number", "%" = "percentage", "'" = "bla", "\"" = "")
  }
  
  snakecase::to_any_case(string, 
                       case = case, 
                       sep_in = "[^[:alnum:]]",
                       parsing_option = 1L,
                       abbreviations = names(transliterations),
                       transliterations = c("Latin-ASCII", transliterations),
                       numerals = "asis",
                       unique_sep = "duplicated",
                       empty_fill = "V"
                      )
}

make_clean_names(string)
# [1] "V"                 "V_duplicated_1"    "V_duplicated_2"    "bla_number"       
# [5] "bla_percentage"    "so_bla_so"         "tatsachlich_liebe"
```

If you think this is interesting, especially the transliterations, just let me know, because then I would look rather soon into solving https://github.com/Tazinho/snakecase/issues/186 and https://github.com/Tazinho/snakecase/issues/181 as these might be dealbreakers.

billdenney:
@Tazinho, thanks for the suggestions!

If we can punt some of the additional work to snakecase, I'd be all for it!  You're correct about the issue with `iconv()` relating to the fact that it is locale-dependent.  And, for general ASCII conversion, I want to make it not locale-dependent.

There are a few features of current `make_clean_names()` that your suggested version does not include such as dropping initial spaces and punctuation, and ensuring that the resulting name is usable without quoting it.  (I.e. being able to use `my_data$clean_name` instead of `` my_data$` clean_name` `` if the input had a name of `" clean name"`-- note the initial space).

While I like transliterations occurring within snakecase, an underlying issue comes with things like an input of `"% decrease"` where if we do the transliteration after dropping initial punctuation and spaces, we would get "decrease" instead of "percent_decrease".  So, I'm not sure that we can move transliterations to snakecase as we have a few different goals here.

Overall, I'm not sure that we can really shorten it as much as you suggest while maintaining all the features that we have.  That said-- any amount we can simplify is good, so let's think about it a bit more.  (Though we have to have it tied up by the 14th, so that CRAN can be appeased.)

Tazinho:
Hmm, I believe these are very strong points that you mention. I always underestimate how tricky this can become.

> If we can punt some of the additional work to snakecase, I'd be all for it! You're correct about the issue with `iconv()` relating to the fact that it is locale-dependent. And, for general ASCII conversion, I want to make it not locale-dependent.

If you have some specific testcases I'd be happy to add them to snakecase. This might be nice if we can get at some point around `make.names()`. (Not sure if this has potential to help then, but maybe yes, I would be happy to try).

> There are a few features of current `make_clean_names()` that your suggested version does not include such as dropping initial spaces and punctuation, and ensuring that the resulting name is usable without quoting it. (I.e. being able to use `my_data$clean_name` instead of `` my_data$` clean_name` `` if the input had a name of `" clean name"`-- note the initial space).

I think when we have `sep_in = "[^[:alnum:]]"` the issue with punctuation in the beginning should be handled. (snakecase trims it's output and the resulting string will contain only alphanumerics, underscores and possibly any unvalid things that the user might supply in transliterations, so one should also add a test to allow only alphanumerics and underscores in the element slots of the transliterations argument). The only problem I see here is that it doesn't safe us from digits in the beginning and reserverd words (`help(Reserved)`) so one still might need some boilerplate code in the end, i.e. like 

```r
string[string != make.names(string)] <- paste("x_", string[string != make.names(string)])
```

The case of `x_` could depend on the supplied value for the `case` argument (and the value `x_` could be itself an argument to `make_clean_names()`).

> While I like transliterations occurring within snakecase, an underlying issue comes with things like an input of `"% decrease"` where if we do the transliteration after dropping initial punctuation and spaces, we would get "decrease" instead of "percent_decrease". So, I'm not sure that we can move transliterations to snakecase as we have a few different goals here.

I am not sure how mature the abbreviations argument in snakecase is but here it works as it should and if we drop the preprocessing step in `make_clean_names()` this would also work in make_clean_names (if this feature is really wanted). See the last element of `string`.

```r
string <- c("", "", "", 
            "bla #", "bla %", "so'so",
            "tatsächlich\"Liebe", "# decrease")
make_clean_names(string)
# [1] "V"                 "V_DUPLICATED_1"    "V_DUPLICATED_2"    "bla_number"       
# [5] "bla_percentage"    "so_bla_so"         "tatsachlich_liebe" "number_decrease" 
```

In summary, I'd change the mainpart of my suggestion to (partly pseudocode)

```r
make_clean_names <- function(string, case = "snake", transliterations = NULL) {
  
  # check if transliterations contains unallowed characters (non alphanumeric or non "_")
  if(!is.null(transliterations) & any(str_detect(transliterations, pattern = "[^[:alnum:]^_]"))) {
    stop("`transliterations` must only contain characters, digits and underscores")
  }
  
  string <- snakecase::to_any_case(string, 
                                   case = case, 
                                   sep_in = "[^[:alnum:]]",
                                   parsing_option = 1L,
                                   abbreviations = names(transliterations),
                                   transliterations = c("Latin-ASCII", transliterations),
                                   numerals = "asis",
                                   # unique_sep = "_duplicated_",
                                   empty_fill = to_any_case("V", case = case)
  )
  
  prefix <- to_any_case("x", case = case, postfix = "_")
  string[string != make.names(string)] <- paste0(prefix, string[string != make.names(string)])
  # As the last line can introduce duplications, I switched the duplications part again out of
  # snakecase::to_any_case(). (See the two last strings in the test case below).
  make.unique(string, sep = "_duplicated_")
}
string <- string <- c("", "", "", 
                      "bla #", "bla %", "so'so",
                      "tatsächlich\"Liebe",
                      "# decrease", "x_123", "123")

make_clean_names(string)
# [1] "v"                  "v_duplicated_1"     "v_duplicated_2"     "bla"               
# [5] "bla_duplicated_1"   "so_so"              "tatsachlich_liebe"  "decrease"          
# [9] "x_123"              "x_123_duplicated_1"
```

> Overall, I'm not sure that we can really shorten it as much as you suggest while maintaining all the features that we have. That said-- any amount we can simplify is good, so let's think about it a bit more. (Though we have to have it tied up by the 14th, so that CRAN can be appeased.)

I am also not sure. Also about the timeline. But if you like the approach and have some feedback I will be happy to put some more time in this (and try to fix some issues in snakecase with higher priority).

Tazinho:
I already see another issue with my suggestion popping up. The `"so'so"` gets split by snakecase (this is correct behaviour which I oversaw) and so, we get an unwanted case conversion here. I think without overthinking this, my suggestion probably doesn't have potential to help ATM.

```
make_clean_names(string, case = "small_camel")
# [1] "v"                "v_duplicated_1"   "v_duplicated_2"   "blaNumber"       
# [5] "blaPercentage"    "soSo"             "tatsachlichLiebe" "numberDecrease"  
# [9] "x123"             "x_123"  
```

R/make_clean_names.R
```diff
@@ -24,58 +51,91 @@
...
+        id="Greek-Latin;Latin-ASCII;Accents-Any;Any-ASCII"
+      )
+    } else {
+      replaced_names
+    }
+  # Remove starting spaces and punctuation
+  good_start <-
+    stringr::str_replace(
+      str=transliterated_names,
+      pattern="\\A[\\h\\s\\p{Punctuation}\\p{Symbol}\\p{Separator}\\p{Other}]*(.*)$",
```
sfirke: I can't sight-read this, it's beyond my regex familiarity.  Can you explain it or provide a reference? :pray:

R/make_clean_names.R
```diff
@@ -24,58 +51,91 @@
...
+      replacement="\\1"
+    )
+  # make.names() is dependent on the locale and therefore will return different
+  # system-dependent values.
```
sfirke: Do you have a sense of, in practice, what the most common locale inconsistency in `make.names` might be?  Just wondering.

R/make_clean_names.R
```diff
@@ -24,58 +51,91 @@
...
-      case = case, sep_in = "\\.",
-      transliterations = c("Latin-ASCII"), parsing_option = 1,
+
+  replaced_names <-
+    stringr::str_replace_all(
+      str=string,
+      pattern=replace
+    )
+  transliterated_names <-
+    if (ascii) {
```
sfirke: This is all so readable :heart_decoration:

R/make_clean_names.R
```diff
@@ -24,58 +51,91 @@
...
+      str=transliterated_names,
+      pattern="\\A[\\h\\s\\p{Punctuation}\\p{Symbol}\\p{Separator}\\p{Other}]*(.*)$",
+      replacement="\\1"
+    )
+  # make.names() is dependent on the locale and therefore will return different
+  # system-dependent values.
+  made_names <- make.names(good_start)
+
+  # Handle dots, multiple underscores, case conversion, string transliteration
+  # Parsing option 4 removes underscores around numbers, #153
```
sfirke: This comment looks obsolete

tests/testthat/test-clean-names.R
```diff
@@ -1,42 +1,197 @@
...
+  )
+  expect_equal(
+    make_clean_names("a/b"),
+    "a_b",
+    info="No custom replacement"
+  )
+  expect_equal(
+    make_clean_names("a/b", replace=c("/"="_per_")),
+    "a_per_b",
+    info="Custom replacement"
```
sfirke: This is great.  Also, what does the `info=` part do?

tests/testthat/test-clean-names.R
```diff
@@ -1,42 +1,197 @@
...
+    "a b c d e f",
+    "testCamelCase",
+    "!leadingpunct",
+    "average # of days",
+    "jan2009sales",
+    "jan 2009 sales",
+    "not_first_unicode_µ",
+    "µ_first_unicode"
+  )
+result_vector_noascii <-
```
sfirke: Are these result vectors used again?  I don't see them

tests/testthat/test-clean-names.R
```diff
@@ -1,42 +1,197 @@
...
+    "a", # multiple special chars, trailing special chars
+    "percent", # converting % to percent
+    "x", # 100% invalid name
+    "x_2", # repeat of invalid name
+    "d_9", # multiple special characters
+    "repeated_2", # uppercase, 2nd instance of repeat
+    "cant", # uppercase, 2nd instance of repeat
+    "hi_there", # double-underscores to single
+    "leading_spaces", # leading spaces
+    "x", # euro sign, converted
```
sfirke: Why is this `x` and not `x_3` like in the no_ascii result vector?

More generally, is there test coverage showing the difference between `ascii = TRUE` and `FALSE`?  That was the only test that seemed missing.  For instance above `ação` turns into `acao` for both `TRUE` and `FALSE`.

tests/testthat/test-clean-names.R
```diff
@@ -1,42 +1,197 @@
...
+    "acao", # accented word, transliterated to latin,
+    "faroe", # Farœ character was failing to convert on Windows, should work universally for stringi 1.1.6 or higher
+    # https://github.com/sfirke/janitor/issues/120#issuecomment-303385418
+    "a_b_c_d_e_f", # for testing alternating cases below with e.g., case = "upper_lower"
+    "test_camel_case", # for testing alternating cases below with e.g., case = "upper_lower"
+    "leadingpunct", # for testing alternating cases below with e.g., case = "upper_lower"
+    "average_number_of_days", # for testing alternating cases below with e.g., case = "upper_lower"
+    "jan2009sales", # no separator around number-word boundary if not existing already
+    "jan_2009_sales", # yes separator around number-word boundary if it existed
+    "not_first_unicode_µ",
```
sfirke: I don't see this vector used, if I did I would expect this mu to become `m`?  And then without the results vectors used, right now there's no test for the preservation of delta or mu when `ascii = TRUE`.  I may just be missing things here, sorry if so.

R/make_clean_names.R
```diff
@@ -24,58 +51,91 @@
...
+                             case = c(
+                               "snake", "lower_camel", "upper_camel", "screaming_snake",
+                               "lower_upper", "upper_lower", "all_caps", "small_camel",
+                               "big_camel", "old_janitor", "parsed", "mixed", "none"
+                             ),
+                             replace=
+                               c(
+                                 "'"="",
+                                 "\""="",
+                                 "%"=".percent_",
```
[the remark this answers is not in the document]
sfirke: I didn't see it clearly in a test, could you add one to show what the `.` preceding does?  Then I'd understand where we need to note anything about that to the user, if they'd want to use the `.` the same way.

NEWS.md
```diff
@@ -1,18 +1,28 @@
 # janitor 1.2.1.9000 (unreleased), will be v 1.3.0
 
+## Breaking Changes
+
+* `clean_names()` and `make_clean_names()` now work significantly harder to be locale-independent, and translation to ASCII is simpler. These changes alter the new names provided by these functions in some cases. (Fix #331, thanks to @billdenney)
```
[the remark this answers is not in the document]
sfirke: For instance we can say how delta now becomes d on any platform.  And let's note the control over substitutions, they can now extend and control `%` -> `percent`.

billdenney:
Other issues that will be resolved with this PR:
Fix #268
Fix #271 
Fix #283

NEWS.md
```diff
@@ -1,18 +1,28 @@
 # janitor 1.2.1.9000 (unreleased), will be v 1.3.0
```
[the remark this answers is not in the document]
billdenney: I'll leave that to another PR when everything is lined up.

NEWS.md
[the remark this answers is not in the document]
billdenney: I simplified this language and added the example for percent, too.  Good call.

NEWS.md
```diff
@@ -60,6 +70,8 @@ This feature (#50) took almost 3 years from conception to implementation.  Major
 
 * If the third variable in a three-way tabyl is a factor, the resulting list is sorted in order of its levels (#250).  Empty factor levels in the 3rd variable are still omitted regardless of the value of `show_missing_levels`.
 
+* `clean_names()` can now be called on *tbl_graph* objects from the `tidygraph` package. (#252, thanks to @gvdr for bringing up the issue and thanks to @Tazinho for proposing solution).
```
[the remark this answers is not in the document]
billdenney: Whoops!  Fixed.

R/clean_names.R
```diff
@@ -39,48 +37,55 @@
...
-  "big_camel", "old_janitor", "parsed", "mixed", "none"
-)) {
-  if(!is.data.frame(dat)){ 
-    stop( "clean_names() must be called on a data.frame.  Consider janitor::make_clean_names() for other cases of manipulating vectors of names.") 
-  }
-  stats::setNames(dat, make_clean_names(names(dat), case = case))
+#' @rdname clean_names
+#' @export
+clean_names.default <- function(dat, ...) {
+  stop(
```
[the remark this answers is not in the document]
billdenney: :)

R/make_clean_names.R
```diff
@@ -1,18 +1,45 @@
...
 #' 
-#' Accented characters are transliterated to ASCII.  For example, an "o" 
-#' with a German umlaut over it becomes "o", and the Spanish character "enye" becomes "n".
+#' When `ascii=TRUE` (the default), accented characters are transliterated to
+#' ASCII.  For example, an "o" with a German umlaut over it becomes "o", and the
+#' Spanish character "enye" becomes "n".
+#' 
+#' The order of operations is: `replace`, (optional) ASCII conversion, removing
+#' initial spaces and punctuation, apply `base::make.names()`, apply
+#' `snakecase::to_any_case()`, and add numeric suffixes to duplicates.
```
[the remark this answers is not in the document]
billdenney: Thanks!

R/make_clean_names.R
```diff
@@ -1,18 +1,45 @@
...
 #' @param string A character vector of names to clean.
-#'
-#' @inheritParams clean_names
+#' @param case The desired target case (default is \code{"snake"}), indicated by
+#'   these possible values:
+#' \itemize{
+#'  \item{\code{"snake"} produces snake_case}
+#'  \item{\code{"lower_camel"} or \code{"small_camel"} produces lowerCamel}
+#'  \item{\code{"upper_camel"} or \code{"big_camel"} produces UpperCamel}
+#'  \item{\code{"screaming_snake"} or \code{"all_caps"} produces ALL_CAPS}
```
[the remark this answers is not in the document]
billdenney: I just linked directly to the snakecase documentation here.

R/make_clean_names.R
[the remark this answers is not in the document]
billdenney: I changed it to "_percent_".  It's just to add the separator (that was how it was before, too).  And, I added a test, too.

R/make_clean_names.R
billdenney: This is not normal regexp-foo.  It's ICU-regexp foo!  The main advantage of ICU regexps are that they use Unicode and are locale-independent.  I've added some documentation for what it all means as comments.

R/make_clean_names.R
```diff
@@ -24,58 +51,91 @@
...
-  "big_camel", "old_janitor", "parsed", "mixed", "none"
-)) {
+#'
+#' @importFrom stringi stri_trans_general
+#' @importFrom stringr str_replace str_replace_all
+make_clean_names <- function(string,
+                             case = c(
+                               "snake", "lower_camel", "upper_camel", "screaming_snake",
+                               "lower_upper", "upper_lower", "all_caps", "small_camel",
+                               "big_camel", "old_janitor", "parsed", "mixed", "none"
```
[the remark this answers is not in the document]
billdenney: This now just goes for everything from `to_any_case()`.

R/make_clean_names.R
billdenney: The issue in #268 is one example.  I think that it mostly relates to Unicode characters that are local letters.

R/make_clean_names.R
billdenney: :)

R/make_clean_names.R
billdenney: Removed.

tests/testthat/test-clean-names.R
billdenney: If you ever do detailed reporting on tests, you can show the info part as part of that reporting.  In most cases, it's probably identical to just putting a comment to describe the test.

tests/testthat/test-clean-names.R
billdenney: I added a test for "μ".  Given all the attempts to convert to something usable, Greek delta gets converted in almost every case while mu only requires `ascii=FALSE`.  I'm not sure of a simple way around the almost-always conversion for delta.

tests/testthat/test-clean-names.R
billdenney: I had intended to use them, but I ended up going down a different path.  I've removed them.

tests/testthat/test-clean-names.R
billdenney: Not being `x_3` is part of the rework where I was having some issues with my testing.  I added a test for `ascii=TRUE`, but the number of ways that transliterations happen makes it difficult to fully control allow for e.g. the acao scenario.

I'm guessing that at the end of this, we'll have some additional work with Unicode characters and other non-ASCII characters, but I think that in the general cases, we're doing better.  (And more to the point, I don't see a way to do all that we are wanting to do without some crunching of unusual characters.)

billdenney:
The tests that are now failing are, I think locale-specific.  That gives me pause for how testable these are overall.

billdenney:
@sfirke, I'm not sure what else I can do for the locale-specific tests that I've now separated out in test-clean-names.R.  If this version doesn't work on Travis, we may need to skip these tests and try to fix them in 2.0.1.

I've spent a good chunk of time going down Unicode/locale rabbit holes and my main take-away is that the interaction of Unicode and locales is a horrible mess (and no two OSes have the same locales available).

sfirke:
You are a saint for digging into that mess!  I will take a look at this in its current state and then yeah, we can always comment out tests failing for locale-specific reasons :grin:

sfirke:
Hey-hey, this version passes on Travis!  :tada:  I'm good with merging it in, please go ahead.

Did you end up withholding any tests?  If there are any tricky ones that you want to retain with this code, you could add them but comment them out.

sfirke:
Not about the actual code: I could see adding a test each for the issues being closed like #268 , #271 , #283 etc. and for the ones that are new features, like #271 and #283 , add that functionality to the examples of `clean_names` and/or `make_clean_names` and describe in NEWS.  So people know about and use the new powers! :person_fencing: 

Let's not have that stop the merging of this PR though, and if you've had enough of `clean_names` the last couple of weeks I could take over this aspect :smile:

billdenney:
I think that I added tests for all but #204, but all to_any_case arguments are now available, so that should work, too.

I will admit, I'm tiring of the morass of Unicode and locales that I've been working through on this.  If you can take on the examples and other parts, that would definitely be appreciated.

Merging now!

Issue #329, opened 2020-02-02
get_dupes fails when called on grouping variables


sfirke:
A rare case, but yes, it happened to me in actual analysis this week and was slightly annoying.
```
> mtcars %>% group_by(cyl) %>% get_dupes(cyl)
Error: Column `cyl` can't be modified because it's a grouping variable
```

I think `get_dupes()` should check if a data.frame is grouped and then ungroup it if so before proceeding.  I don't see a meaningful way in which a user would expect grouping to have a practical effect and anyway, it doesn't so they'd be mislead.

`mtcars %>% group_by(cyl) %>% get_dupes(mpg)` is okay, I think the grouping is meaningless.  In short, `get_dupes()` is an interactive function and it shouldn't choke for this unrelated reason.

jzadra:
I'll take this one.

billdenney:
I think that an ungrouping warning should be given just in case the user was expecting something else (like `get_dupes()` to be run per group level).

jzadra:
> I think that an ungrouping warning should be given just in case the user was expecting something else (like `get_dupes()` to be run per group level).

To me this seems like a slightly separate issue - running get_dupes on given vars that are NOT the grouping vars, the user might expect get_dupes to only consider within each group.  However Sam's issue was more about calling it on a grouping variable itself, in which case there is no need to consider the grouping variable.  Of course doing a combination of the two requires some consideration.

jzadra:
But to fix the failure, something like this would do the trick:

```
x <- mtcars %>% group_by(cyl, gear)



if is_grouped_df(x) {
  
xgroups <- group_vars(x)

x <- x %>% ungroup()

}
  
## standard get dupes function here ##

if exists("xgroups") {
x %>% 
  group_by(!!!syms(xg))

  }
```

billdenney:
Ah, got it.  I agree that what I mentioned would be a different issue.

sfirke:
Yep I was just talking about the simple failure when it is called on a
grouping variable. Let's go ahead with that.

Anyhow, after thinking some, I don't know what other behavior a user might
expect the grouping attribute to cause.

sfirke:
[...]

PR #345, opened 2020-03-12
get_dupes() grouped data error fix

jzadra:
## Description

If data is grouped:
1. Saves the grouping structure and then removes it before checking for duplicates. 
2. One warning per session is issued notifying user that get_dupes() isn't checking within groups but rather over the whole data set, however will still return an object with the same grouping structure.
3. Groups are reapplied before returning dupes.

Added tests that check that

1. The grouping structure of the original data is the same as the result.
2. The output for grouped data is the same as the output for ungrouped data other than the grouping structure.

## Related Issue
Addresses #329.

## Example
x <- mtcars %>% group_by(cyl, gear)
get_dupes(x)

### DIFF
--- a/R/get_dupes.R
+++ b/R/get_dupes.R
@@ -24,6 +24,15 @@ get_dupes <- function(dat, ...) {
   
   names(dat)[pos] <- names(pos) #allows for renaming within get_dupes() consistent with select()
   
+  #Check if dat is grouped and if so, save structure and ungroup temporarily
+  is_grouped <- dplyr::is_grouped_df(dat)
+  
+  if(is_grouped) {
+    dat_groups <- dplyr::group_vars(dat)
+    dat <- dat %>% dplyr::ungroup()
+    message(paste0("Data is grouped by [", paste(dat_groups, collapse = "|"), "]. Note that get_dupes() is not group aware and does not limit duplicate detection to within-groups, but rather checks over the entire data frame. However grouping structure is preserved."))
+  }
+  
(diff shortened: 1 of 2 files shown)


jzadra:
I'm not clear on exactly what codecov is if anyone can provide some feedback on what I've done wrong!

sfirke:
codecov checks which lines of code are run when all the tests are run.  Your PR has two "misses", that is there are two lines that are not run in any test: https://codecov.io/gh/sfirke/janitor/pull/345/diff?src=pr&el=tree#diff-Ui9nZXRfZHVwZXMuUg  Looks like it's the grouped_df warning, that could be checked with a new test that looks at `expect_message`.

sfirke:
Code itself looks great though! 👍 After reviewing it, now I wonder if it's impossible to test that chunk that codecov says is missing coverage, because the message only prints if `interactive()` and in the test environment `interactive()` returns false.

I don't see a way to fake an interactive environment to run that warning.  The only idea I have is to just not have tests for those two lines and acknowledge that by ending those two lines with `# nocov` to tell codecov to ignore them.

jzadra:
Looks like you were correct about the interactive issue.  I'll add #nocov to those lines.

sfirke:
Looks like it didn't run, I'm guessing because `interactive()` would be FALSE?  I think you can just add `# nocov to those two lines in the function itself.

sfirke:
Excellent!  Thanks @jzadra

Issue #362, opened 2020-04-07
easy Suggestion for janitor: adorn_totals("both")  # shows row and col totals...


sfd99:
GREAT, Sam!
janitor is wonderful.

btw:
a **shortcut** to get Total Sums 
for _BOTH_ rows _AND_ cols:

mtcars %>%
tabyl(am, cyl) %>%
adorn_totals(c("row", "col"))
am 4 6 8 Total
0 3 4 12 19
1 8 3 2 13
Total 11 7 14 32

So, 
 (easy) SUGGESTION -
also allow keyword:
"**_both_**"
as param to:
adorn_totals("**_both_**")
or maybe simply:
  adorn_totals()

less coding...easier... :-)

higgi13425:
kinda like this one. If it is not too hard to do.

sfd99:
excellent, Sam,
hey! looking forward to it.
less keystrokes the better... :-)
**Janitor** pkg. is _so super- nice_to use...useful!!

jzadra:
This is far simpler, but some of the discussion in #313 might be of use.

svgsstats:
I like this suggestion as well. 

I use this package in my Intro Stats course and we bring it into the course early on.  The current method means that I need to have some explanation of vectors which isn't necessary (or desired early on).

sfirke:
Okay, let's do it, with `where = "both"` being the equivalent to `where = c("col", "row")`.  I think `row` should probably remain the default, (a) since that's less breakage (b) only `row` makes sense on a one-way tabyl.

This is relatively basic, if anyone wants to submit a pull request with this change + add it to NEWS + add a test, I'd welcome it!

PR #403, opened 2020-09-13
changed adorn_totals to take where = both argument

svgsstats:
Changed the adorn_totals() function so that where = "both" is equivalent to where = c("row","col")

Added a test as well.

## Description
Just added a simple conditional to the adorn_totals() function.  

I didn't update the NEWS.md file because I was unsure of the way that should get done.

## Related Issue
improvement as per #362

## Example
ct %>% tabyl(where="both")

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # janitor 2.0.1.9000 (unreleased)
 
+## New features
+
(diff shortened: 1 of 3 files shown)


sfirke:
Looks great, thanks! I'll update NEWS and merge later.

sfirke:
I think we're set but [Travis is being fussy](https://travis-ci.com/github/sfirke/janitor/builds/184403579), I'm not sure why the builds (for master and thus this branch) are failing.  Usually it resolves itself, so I'll give it a few days, then merge (I hope).  Thanks again!

sfirke:
I can't figure out what's happening with one particular Travis build combo but clearly this is not the cause and it's passing on the other builds so I'm going to merge it.  Thank you again!

PR #404, opened 2020-10-12
Added option to pass NA to fill argument of adorn_totals to preserve column types

### LINKED ISSUE #150: Add test for print.tabyl
To confirm it prints w/o row numbers.  See https://stackoverflow.com/q/47397008/4470365

### LINKED ISSUE #298: Feature request: adorn_totals for tibbles
Adorn_totals is a nice function. But when it's run on a tibble it changes the data type, and information is lost.

hamstr147:
## Description

I have implemented a type check in the `col_sum` function in `adorn_totals` such that the function checks the data type of a column and uses the relevant `NA` variant for the total column. This only happens if the user does not supply a string (e.g. "-" as the fill value - I have kept this as the default). This also required some changes in the subsequent `purrr::map_df` call, but I have kept the original implementation as the default behavior. 

I have also updated the documentation to reflect this change and written a unit test in the `test-add-totals.R` script. Note that I have **not** updated the NEWS file as I don't know if there is a preferred wording/style so will leave that up to the maintainer.

## Related Issue

fixes #298. 

## Example

```
test_df <- data.frame(
    a = c("hi", "low", "med"),
    b = factor(c("big", "small", "regular")),
    c = c(as.Date("2000-01-01"), as.Date("2000-01-02"), as.Date("2000-01-03")),
    d = 1:3,
    e = 4:6,
    f = c(TRUE, FALSE, TRUE),
    g = c(7.2, 8.2, 9.2),
    stringsAsFactors = FALSE
  )
  
  out <- adorn_totals(test_df, fill = NA)
```
This produces a `tabyl` with a total row that preserves the data types for each column:
```
out

     a       b          c d  e     f    g
    hi     big 2000-01-01 1  4  TRUE  7.2
   low   small 2000-01-02 2  5 FALSE  8.2
   med regular 2000-01-03 3  6  TRUE  9.2
 Total    <NA>       <NA> 6 15    NA 24.6

str(out)

Classes ‘tabyl’ and 'data.frame':	4 obs. of  7 variables:
 $ a: chr  "hi" "low" "med" "Total"
 $ b: Factor w/ 3 levels "big","regular",..: 1 3 2 NA
 $ c: Date, format:  ...
 $ d: int  1 2 3 6
 $ e: int  4 5 6 15
 $ f: logi  TRUE FALSE TRUE NA
 $ g: num  7.2 8.2 9.2 24.6
 - attr(*, "core")='data.frame':	3 obs. of  7 variables:
  ..$ a: chr [1:3] "hi" "low" "med"
  ..$ b: Factor w/ 3 levels "big","regular",..: 1 3 2
  ..$ c: Date[1:3], format:  ...
  ..$ d: int [1:3] 1 2 3
  ..$ e: int [1:3] 4 5 6
  ..$ f: logi [1:3] TRUE FALSE TRUE
  ..$ g: num [1:3] 7.2 8.2 9.2
 - attr(*, "tabyl_type")= chr "two_way"
 - attr(*, "totals")= chr "row"
```

### DIFF
--- a/R/adorn_totals.R
+++ b/R/adorn_totals.R
@@ -5,7 +5,7 @@
 #'
 #' @param dat an input data.frame with at least one numeric column.  If given a list of data.frames, this function will apply itself to each data.frame in the list (designed for 3-way \code{tabyl} lists).
 #' @param where one of "row", "col", or \code{c("row", "col")}
-#' @param fill if there are non-numeric columns, what string should fill the bottom row of those columns?
+#' @param fill if there are non-numeric columns, what should fill the bottom row of those columns? If a string, relevant columns will be coerced to character. If `NA` then column types are preserved. 
 #' @param na.rm should missing values (including NaN) be omitted from the calculations?
 #' @param name name of the totals column or row 
 #' @param ... columns to total.  This takes a tidyselect specification.  By default, all numeric columns (besides the initial column, if numeric) are included in the totals, but this allows you to manually specify which columns should be included, for use on a data.frame that does not result from a call to \code{tabyl}. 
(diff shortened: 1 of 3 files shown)


sfirke:
This looks great, thank you! 🙌 

Travis is having random time-outs on trying to build the package, ideally if I re-run it a few times there it'll pass and then I'll squash and merge.  If not I'll merge anyway 😁   And I'll add your contribution to the NEWS page.  Thank you for taking this feature on and doing it well.

sfirke:
The Travis problem persists on OS X only because a font won't install...  I'm merging anyway and if it fails on master, 🤷

sfirke:
thanks again!

Issue #359, opened 2020-04-06
adorn_totals to accept a vector for name argument when adding row and column totals


francisbarton:
## Feature request

I wanted to add totals to rows and columns, but to the give them different names, like this:

```
janitor::adorn_totals(where=c("row", "col"), name = c("bracket_total", "area_total"))

Error: Assigned data `name` must be compatible with row subscript `1`.
x 1 row must be assigned.
x Assigned data has 2 rows.
i Only vectors of size 1 are recycled.
```

I think this would be a good feature, when `where` takes both "row" and "col", `name` should also be able to take a vector. Thanks!

sfirke:
Great idea!

francisbarton:
Thank you! I use {janitor} all the time, by the way.
I haven't had a look at the code at all yet - and I am not an experienced programmer - but is this FR something you think you can do relatively easily, or would you welcome me having a crack at an implementation?

sfirke:
Ordinarily I'd say this is a good place to start off. I am trying to move
pretty quickly these days because I'm overdue on submitting this version to
CRAN. On the other hand, I have very little time and can use all the help I
can get.

I appreciate your interest in helping. For starters, do you want to take a
look at the code for adorn_totals? From there I think it might be clear to
you when you play around with the it whether you can make this behavior
work as you intend. If it seems solvable to you, then why don't you fork
this repository, make the changes you can, and submit a pull request and I
will work with you on it. No worries if that feels like too much. And if
you do take the plunge, I can answer questions.

francisbarton:
Yup - I'll have a look and maybe fork and see what I can do. Thank you for the offer of support with it.

sfirke:
@francisbarton do you have interest in working on this?  Could be nice to get it into the next release of janitor

francisbarton:
Hi Sam,
Yes I am interested in this, I just don't know if I have the capability to make the changes. When I looked at it before, I got a bit lost in how it all worked. But that was several months ago. I'll have another look now and get back to you.

sfirke:
closed by #413

PR #413, opened 2020-11-27
Enable passing of a length 2 vector as name argument (addresses #359)

francisbarton:
## Description
Allow the `name` argument to accept a vector, as well as a single value (as currently).

If a vector of length 2 is supplied, and `where` is equal to `c("row", "col")`, the first element will be used as the first column in the appended totals row (unless column 1 is included in cols to be totalled), and the second element will be used as the column name in the appended totals column.
This allows the user to give different headers to a totals row and a totals column, where previously these had to have the same value.

If a single item is supplied, this is duplicated into a vector of length 2. This will supply the same name for the row and column headers, as above - and providing the same functionality as previously - if both "row" and "col" are stipulated.

If only one of "row" or "col" is stipulated, it still works, as the function will just take either the first or the second element as necessary (and both are identical).

If only one of "row" or "col" is stipulated, but the user in error provides a longer vector for `name`, the vector is trimmed back leaving just the first element.

I have also included suggested revisions (commented out) to the add totals row and add totals column sections, to use a pure dplyr process. But it may be that these result in forcing tibble output rather than data.frame output, so please ignore and delete suggested lines if unwanted.

## Related Issue
Addresses feature request in #359

## Example
No change to existing usage; pure addition of extra option.

No tests included.

PR includes 4 previous commits which were just bringing the fork up to date. The change to `adorn_totals.R` is only in one of the commits, but I don't know how to remove the other 4 from the PR.


francisbarton:
Looks like a good opportunity for me to practise writing tests!

sfirke:
I just skimmed it all and it looks good!  Thank you.  I will find some time in the next few days to review and run line by line, then make any edits if needed and merge.  Only thing left I see immediately is to add this update to NEWS.Md, thanking you for the contribution - feel free to add that if you like :grin:

R/adorn_totals.R
```diff
@@ -133,9 +144,28 @@ adorn_totals <- function(dat, where = "row", fill = "-", na.rm = TRUE, name = "T
...
+    #     dplyr::bind_cols(dat, .)
+    # }
+    
+    # dplyr >= 1.0 version:
+    # if ("col" %in% where) {
+    #   # Add totals col
+    #   dat <- dat %>%
+    #     dplyr::rowwise() %>% 
+    #     dplyr::mutate(name[2] = sum(c_across(all_of(cols_to_total) & where(is.numeric)), na.rm = na.rm)) %>% 
+    #     dplyr::ungroup()
```
sfirke, line 166: Thanks for these!  I tried the dplyr >= 1.0 one for a while but had failing tests so just for the sake of time I kept what I had.  I got as far as changing your line 165 to `mutate("{name[2]}" := sum(c_across` - the `mutate(name[2]` wasn't working for me.

sfirke: No need to work on it, I think status quo is okay - just sharing my experience in case we come back to this someday.

sfirke:
hey @francisbarton I am having Git trouble and ended up with #414, a separate pull request.  Still has your commits though!  I will probably merge that one and close this one.

PR #421, opened 2021-01-15
Warn for daylight savings time missing hours

### LINKED ISSUE #420: excel_numeric_to_date gives NA output for times in daylight savings time gap
The function excel_numeric_to_date seems to work fine with most input, but fails on others when include_time is specified.

Example:  
> excel_numeric_to_date(43170.08, include_time=TRUE)
[1] "2018-03-11 01:55:12 PST"
> excel_numeric_to_date(43170.09, include_time=TRUE)
[1] NA
> excel_numeric_to_date(43170.10, include_time=TRUE)
[1] NA
> excel_numeric_to_date(43170.11, include_time=TRUE)
[1] NA
> excel_numeric_to_date(43170.12, include_time=TRUE)
[1] NA
> excel_numeric_to_date(43170.13, include_time=TRUE)
[1] "2018-03-11 03:07:12 PDT"

Most of the values where the function fails seem to end in .09 through 0.11 range.

billdenney:
Fix #420

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 # janitor 2.1.0.9000 (unreleased, under development)
 
+## Minor features
 
+* `excel_numeric_to_date()` now warns when times are converted to `NA` due to hours that do not exist because of daylight savings time.  (fix #420, thanks **@Geomorph2** for reporting and **@billdenney** for fixing)
 
 # janitor 2.1.0 (2021-01-05)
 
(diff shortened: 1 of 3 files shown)


billdenney:
Hmm, Windows and Mac work while Linux does not.  This will take some investigation.

billdenney:
This is included in #422 , so working on it there, only.

PR #425, opened 2021-02-05
remove_empty() example with blanks and NAs

### LINKED ISSUE #424: count rows as empty if entries are NA *or* blank
It would be nice to have an option in `remove_empty()` to remove rows that are composed entirely of *either* `NA` values or are zero-length (e.g. empty strings). (Possibly consider all-whitespace strings to be "empty" as well?)   This is a fairly common pattern in data frames read in from Excel.

The example below isn't particularly "tidy", nor entirely integrated with the `remove_empty()` machinery, but I think shows what I have in mind.

```r
library(tidyverse)
library(janitor)

dd <- tibble(x=c(LETTERS[1:5],rep("",2)),
             y=c(1:5,rep(NA,2)))
remove_empty(dd,"rows")      ## doesn't remove rows 6 &  7 (they're not all-NA)

emptyrows2 <- function(x) {
   apply(x,1,
   function(x) {
        y <- unlist(x)
        all(is.na(y) | nchar(y)==0)
    })
}

dd <- dd[!emptyrows2(dd),]
```

bbolker:
## Description

The only change is to add an example for `remove_empty()` that illustrates how to drop rows containing both blank strings and `NA` values.  The only potential unwanted side-effect/consequence I can think of is that the example loads the `dplyr` package (this is already in `Imports:`; I could use `dplyr::` throughout, but that would be unwieldy especially with piped workflows ...

## Related Issue

fixes #424

### DIFF
--- a/R/remove_empties.R
+++ b/R/remove_empties.R
@@ -16,6 +16,21 @@
 #' @examples
 #' # not run:
 #' # dat %>% remove_empty("rows")
+#' # addressing a common untidy-data scenario where we have a mixture of
+#' # blank values in some (character) columns and NAs in others:
+#' library(dplyr)
+#' dd <- tibble(x=c(LETTERS[1:5],NA,rep("",2)),
+#'              y=c(1:5,rep(NA,3)))
+#' # remove_empty() drops row 5 (all NA) but not 6 and 7 (blanks + NAs)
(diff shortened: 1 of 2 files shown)


billdenney (approved):

billdenney:
Thanks!

PR #426, opened 2021-02-05
Warn for dates before 1 in `excel_numeric_to_date()`

### LINKED ISSUE #423: Excel dates before the era begins do not exist (but they do with excel_date_to_numeric())
## Feature requests

It is unclear the best way to handle negative dates in `excel_date_to_numeric()` as they are undefined in Excel.

## Bug reports

In Excel, the formula `=DATE(1899, 12, 31)` shows a date of 1 March 3700.  Overall, it appears impossible to represent dates as dates before the year 1900 in Excel.

The below is not accurate in Excel, but it is rational based on continuing the number line to be negative.  My first thought is to just leave it as is, and my second thought is to issue a warning.

```r
excel_numeric_to_date(-1:0)
# [1] "1899-12-30" "1899-12-31"
```

Thoughts?

Related to #422

billdenney:
Fix #423 

I did not make the warning a once-per-session warning as it likely indicates some form of data input issue that the user would want to fix.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,7 +7,7 @@
 
 ## Minor features
 
-* `excel_numeric_to_date()` now warns when times are converted to `NA` due to hours that do not exist because of daylight savings time.  (fix #420, thanks **@Geomorph2** for reporting and **@billdenney** for fixing)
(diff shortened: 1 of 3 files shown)


sfirke:
Deleting master auto-closed all open PRs?  Yikes.

billdenney:
Yeah, I think that it makes sense to change the name, but it does close the PRs.  (I need to change the name on several of mine, too.)

PR #430, opened 2021-02-09
Add a find_header() function to help with row_to_names()

### LINKED ISSUE #429: Option for row_to_names to find the first complete row of names
## Feature request

I am working with data that looks like this: 

image

And want to get the `row.names` from line 160.  It would be nice if I could tell `row_to_names` to use the first complete row for the names.  Often the clutter above the row names is descriptive text that only occupies some columns, so would get skipped in favor of the first row with no NA values.

I'm not sure how to work it in - it could be:
- Default behavior if no `row_number` is specified, printing a message of "no row number specified, using the first complete row (row X)`
- `row_number` could accept "first_complete", or `0` as a value requesting this behavior.

I prefer the second bullet, though I'm not sure if `first_complete` or `0` is better.  Either way it would be documented.

billdenney:
Fix #429


sfirke:
Awesome!  I should be able to review by end of this week.

sfirke:
Nuts sorry about this.  If you revive it with main branch (hopefully not too onerous?) I can take a look.

PR #443, opened 2021-04-06
Add a find_header() function to help with row_to_names()

### LINKED ISSUE #429: Option for row_to_names to find the first complete row of names

billdenney:
Fix #429

Replaces #430


billdenney:
@sfirke Ping on this one.

sfirke:
Thanks for the nudge, I will review this by end of the month as well.  I am excited about this one!

R/row_to_names.R
```diff
@@ -1,21 +1,48 @@
...
+#'
+#' x %>%
+#'   row_to_names(row_number = "find_header")
 #' @export
-row_to_names <- function(dat, row_number, remove_row = TRUE, remove_rows_above = TRUE) {
+row_to_names <- function(dat, row_number, ..., remove_row = TRUE, remove_rows_above = TRUE) {
   # Check inputs
-  if (length(row_number) != 1 | !is.numeric(row_number)) {
-    stop("row_number must be a numeric of length 1")
+  if (!is.logical(remove_row) & length(remove_row) == 1) {
```
sfirke, line 27: Should this be `if (!(is.logical(remove_row) & length(remove_row) == 1)) {` ?

R/row_to_names.R
```diff
@@ -1,21 +1,48 @@
...
+    stop("remove_row must be a logical scalar, not ", as.character(remove_row))
+  } else if (!is.logical(remove_rows_above) & length(remove_rows_above) == 1) {
+    stop("remove_rows_above must be a logical scalar, not ", as.character(remove_rows_above))
```
sfirke, line 30: Same here as my first comment above about making it `if (!(...` , do we want to fail if either condition isn't met - in which case wrap them in parentheses?

R/row_to_names.R
```diff
@@ -1,21 +1,48 @@
...
+#' x %>%
+#'   row_to_names(row_number = "find_header")
 #' @export
-row_to_names <- function(dat, row_number, remove_row = TRUE, remove_rows_above = TRUE) {
+row_to_names <- function(dat, row_number, ..., remove_row = TRUE, remove_rows_above = TRUE) {
   # Check inputs
-  if (length(row_number) != 1 | !is.numeric(row_number)) {
-    stop("row_number must be a numeric of length 1")
+  if (!is.logical(remove_row) & length(remove_row) == 1) {
+    stop("remove_row must be a logical scalar, not ", as.character(remove_row))
```
sfirke, line 28: Would it be clearer to people who don't know what a scalar is to say "`must be either TRUE or FALSE`?"  Also applies to subsequent use of "logical scalar"

R/row_to_names.R
```diff
@@ -40,3 +67,64 @@ row_to_names <- function(dat, row_number, remove_row = TRUE, remove_rows_above =
...
+#' Find the header row in a data.frame
+#' 
+#' @details
+#' If \code{...} is missing, then the first row with no missing values is used.
+#' If \code{...} has a single character argument, then the first column is
+#' searched for that value.  If \code{...} has a named numeric argument, then
+#' the value of the argument is searched for the name (see the examples).  If
+#' more than one row is found matching a value that is searched for, the first
+#' matching row will be returned (with a warning).
+#'
```
sfirke, line 80: Maybe we note that when searching for a string, the complete row requirement no longer applies.  We could add in a new short paragraph here:

`When searching for a specified column name, the first row with a match will be returned, regardless of the completeness of the rest of that row.`

Is that an improvement, or just more wordy and confusing?  It might be able to replace the last sentence in the preceding paragraph about what happens with multiple matches.

tests/testthat/test-row-to-names.R
```diff
@@ -88,3 +123,91 @@ test_that("row_to_names works on matrices (Fix #320)", {
...
+  )
+  expect_equal(
+    row_to_names(dat=find_correct, row_number="find_header", "D"),
+    setNames(find_correct[4:nrow(find_correct),], c("D", "E"))
+  )
+  expect_equal(
+    row_to_names(dat=find_correct, row_number="find_header", "E"=2),
+    setNames(find_correct[4:nrow(find_correct),], c("D", "E"))
+  )
+})
```
sfirke, line 213: I reviewed all of these tests, they look good

sfirke (commented):
Looks great!  I had a few questions / small suggestions.  Like usual about documentation and error catching, the code itself is 💯

Sorry I took so long to get to this, it's an outstanding effort and will be a very slick function to add to the janitor cart.

sfirke:
I fixed a conflict in `NEWS` introduced by edits I'd made on the main branch in the meantime, so this is again merge-able.

R/row_to_names.R
```diff
@@ -40,3 +67,64 @@ row_to_names <- function(dat, row_number, remove_row = TRUE, remove_rows_above =
...
+#' @family Set names
+#' @examples
+#' # the first row
+#' find_header(data.frame(A="B"))
+#' # the second row
+#' find_header(data.frame(A=c(NA, "B")))
+#' # the second row since the first has an empty value
+#' find_header(data.frame(A=c(NA, "B"), B=c("C", "D")))
+#' # The second row because the second column was searched for the text "D"
+#' find_header(data.frame(A=c(NA, "B", "C", "D"), B=c("C", "D", "E", "F")), "D"=2)
```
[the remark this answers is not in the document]
billdenney: Good point, I updated the example, and I added that as a test.

R/row_to_names.R
billdenney: Good catch.  Updated.

R/row_to_names.R
billdenney: Updated as above.

R/row_to_names.R
billdenney: That's a good clarification.  I added it with a minor modification.

billdenney:
Thanks for the code review and the NEWS merge.  I have addressed your review comments in an updated commit.

sfirke (approved):
Let's ship it!

PR #454, opened 2021-08-04
Class warnings and test that classed warnings can be silenced

### LINKED ISSUE #452: suggestion: option to silence warnings in row_to_names
Could you consider adding an optional parameter in `row_to_names` to not warn about
"Row X does not provide unique names. Consider running clean_names() after row_to_names()."
Something like `warnings = FALSE` or `quiet = TRUE`.

The warning text is displayed always, even if the next line is exactly 
```r
%>%
    janitor::clean_names()
```

In my case, I have a script with many thousands of lines where the warnings (if any) should be thoroughly reviewed.
So this one is inadecuate.

billdenney:
fix #452

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,7 @@
 
 ## Minor features
 
+* Some warning messages now have classes so that they can be specifically suppressed with suppressWarnings(..., class="the_class_to_suppress").  To find the class of a warning you typically must look at the code where the error is occurring.  (#452, thanks to **@mgacc0** for suggesting and **@billdenney** for fixing)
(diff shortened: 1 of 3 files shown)


billdenney:
One limitation of this PR is that I don't know how a user can get the class of a warning without looking at the code.

sfirke:
Looks good, thank you.  This is a creative approach I haven't seen taken elsewhere.

I'm not sure if there's a better way to note the warning type... maybe at the end of the function's documentation, a sentence like "to suppress the warning from this function, use `suppressWarnings(class = "janitor_warn_row_to_names_not_unique")`".

Issue #183, opened 2018-02-20
weights for tabyl


tklebel:
## Feature requests: weights
It would be great, if one could specify weights when using `tabyl`. For survey data weights are very common, without them using `tabyl` does not make much sense in this case.

If the implementation is more or less straightforward and within the scope of your package, I would be happy to assist with a pull. I looked at the source for `tabyl` and it seems to me that passing down an argument to `dplyr::count` for `wt` like the following should be enough?

``` r
library(dplyr)

test_df <- tribble(~x, ~wt,
                   "a", 1,
                   "a", 1,
                   "b", .5,
                   "b", .5)

test_df %>% 
  count(x)
#> # A tibble: 2 x 2
#>   x         n
#>   <chr> <int>
#> 1 a         2
#> 2 b         2

test_df %>% 
  count(x, wt = wt)
#> # A tibble: 2 x 2
#>   x         n
#>   <chr> <dbl>
#> 1 a      2.00
#> 2 b      1.00
```

Or is there something I am overlooking?

sfirke:
Thanks for this clearly stated feature request!  I have been thinking about it and feel unsure.  I see questions of fit and implementation.

**Fit**
Someone else has mentioned wanting this before and it seems like a useful feature to some users.  Janitor's boundaries aren't very crisp, but while tabyl could be seen as a data cleaning tool (e.g., exploring a variable), this starts to be more of a purely data analysis feature.  I don't know of a tidy tools package for survey data specifically, which might be the perfect home for this.

**Implementation**
You're right about modifying `dplyr::count` to use the `wt` argument.  This change would be simple in janitor 0.3.1 but more complex in the new version, because tabyl now includes the former function `crosstab` and a 3-way version as well.  Do/would you use weighting in those contexts, that is, counts of 2-3 variables with weighting by another one?  So say with dplyr, `count(mtcars, cyl, am, wt = mpg)`.  I ask genuinely and would love to hear from anyone on this, as I don't use weighting like this in my own analyses.

If it makes sense and would be used in the 2- and 3- variable contexts, then I think it's implementable.  It's yet another argument to tabyl but that's an acceptable trade-off if enough people would use it.

Would love to hear from other users and janitor contributors/stakeholders!

tklebel:
I have to admit, I am not an expert when it comes to survey weights, but from what I gather, the point is as follows:

[...]

PR #455, opened 2021-08-30
Fix checking for arguments in tabyl.data.frame

gdutz:
I am pretty sure it should check if var1 and var2 are not missing, instead of checking twice for var1.

It stills works at the moment because the check in line 149 filters correctly for one-way tables beforehand.


sfirke:
Indeed, good catch and thank you for the PR!  I am intrigued that you found that error.  Was it causing a problem?  Up until this point I wasn't sure anyone else besides me had ever looked at the internals of `tabyl` 😅

gdutz:
I was looking into implementing weights (see #183) and when trying a few things out I stumbled over it. But it caused no problems for me so far.

Issue #448, opened 2021-07-09
avoid collision of µg and mg becoming mg in `make_clean_names`


IndrajeetPatil:
This harmless change while renaming columns created quite the storm in a teacup for my analysis, so thought I would check if this is the expected behavior or if there is any way for the function to anticipate measurement units and rename accordingly:

``` r
# µg: micrograms, mg: milligrams
janitor::make_clean_names(c("x_µg", "y_mg"))
#> [1] "x_mg" "y_mg"
```

<sup>Created on 2021-07-09 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

P.S. I am showing only two names for a minimal reprex and my actual dataframe is much more complex and with a ton of variables, so renaming columns before using `janitor` is possible but quite cumbersome. 

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value                       
#>  version  R version 4.1.0 (2021-05-18)
#>  os       macOS Mojave 10.14.6        
#>  system   x86_64, darwin17.0          
#>  ui       X11                         
#>  language (EN)                        
#>  collate  en_US.UTF-8                 
#>  ctype    en_US.UTF-8                 
#>  tz       Europe/Berlin               
#>  date     2021-07-09                  
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version    date       lib source                       
#>  assertthat    0.2.1      2019-03-21 [1] CRAN (R 4.1.0)               
#>  backports     1.2.1      2020-12-09 [1] CRAN (R 4.1.0)               
#>  cli           3.0.0      2021-06-30 [1] CRAN (R 4.1.0)               
#>  crayon        1.4.1      2021-02-08 [1] CRAN (R 4.1.0)               
#>  DBI           1.1.1      2021-01-15 [1] CRAN (R 4.1.0)               
#>  digest        0.6.27     2020-10-24 [1] CRAN (R 4.1.0)               
#>  dplyr         1.0.7      2021-06-18 [1] CRAN (R 4.1.0)               
#>  ellipsis      0.3.2      2021-04-29 [1] CRAN (R 4.1.0)               
#>  evaluate      0.14       2019-05-28 [1] CRAN (R 4.1.0)               
#>  fansi         0.5.0      2021-05-25 [1] CRAN (R 4.1.0)               
[...]

PR #457, opened 2021-08-31
Additional documentation of mu/micro conversion

billdenney:
The issue in #448 is occurring for me a lot more.  I've added some additional documentation to assist with the conversion to "u" instead of "m".


sfirke:
Looks good!  This seems kind of long:
```
replace=c("\u00b5"="u",
#' "\u03bc"="u", "\u3382"="u", "\u338c"="u", "\u338d"="u", "\u3395"="u",
#' "\u339b"="u", "\u33b2"="u", "\u33b6"="u", "\u33bc"="u")
```
Do you think it's worth writing our own shortcut for this and including it in janitor + documenting here?  E.g., the above named vector could be called `mu_to_u` and included in the package as a non-exported constant?  I could see it adding a little confusion to how _other_ replacements work, but it also would be a nice shorthand.

I don't feel strongly either way.

billdenney:
That is a **_great_** idea!  I updated the PR.

To work around an unusual name-setting issue in R (with a question on R-package-devel), I created the constant using `setNames()` rather than directly using a named character vector.

sfirke (approved):
Looks great!

sfirke:
Nice implementation, as always 🙌  This is a slick addition.  It's funny, I thought clean_names was "done" years ago.  Then it got a huge refactor for v2.0.  And stuff like this keeps making it better.

PR #465, opened 2021-12-11
Add cutoff argument to remove_empty() (fix #446)

### LINKED ISSUE #446: Feature suggestion: function to remove columns based on missing percentage
I have some large tables that have say, 99.8 % missing.  I'm wondering if it would be useful to have a function similar to `remove_empty()` that would remove columns based on a cutoff of percent missing?

ie `remove_missing_cutoff(df, cutoff = .99)`

billdenney:
Add cutoff argument to remove_empty() (fix #446)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -35,7 +35,7 @@ Imports:
     tidyr (>= 0.7.0)
 License: MIT + file LICENSE
 LazyData: true
-RoxygenNote: 7.1.1
+RoxygenNote: 7.1.2
 Suggests:
     knitr,
     rmarkdown,
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,7 +14,7 @@
 
 * If a `tabyl()` or similar data.frame is sorted (e.g., with `dplyr::arrange()`), then has `adorn_totals()` and/or `adorn_percentages()` called on it, followed by `adorn_ns()`, the Ns will be sorted correctly to match the tabyl they're being adorned on. (fix #407)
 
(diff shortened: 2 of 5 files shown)


sfirke:
Looks great!

Only possibility I see for improvement is in clarifying whether the cutoff is `=`, or `<=`  Say `cutoff = 0.90` -- `(rowSums(is.na(dat))/ncol(dat)) < cutoff` makes it look like a 90% empty row won't be removed, which is also what is implied in the NEWS update.  But then this in the variable description `What fraction (>0 to <=1) of rows or columns must be empty to be removed?` makes it sound like `=`.  I think it can't hurt to be as explicit as possible about whether, if cutoff = 0.9, a column that's 90% empty will be removed or not.

Thank you for adding this! 🙏

billdenney:
Fair point that this should be explicit.  My goal is that it is consistent if we applied the same definition to `cutoff=1`.  (I'm not using the same code to prevent even a possibility of a floating point issue at `cutoff=1` since that is the common/default case.)

You're right that, with cutoff=0.9, a 90% empty column should be removed.  Fix coming momentarily...

sfirke:
Lovely adjustment!

I'd caught that you were using different code for `cutoff == 1` but it hadn't occurred to me that it was to avoid floating point errors.  That saves what could be a very painful floating point precision error for someone 😬 

Thanks again!  Merging now.

billdenney:
I don't know if the floating point catch is required, but it seems like a small code maintenance price to pay for ensuring that it's never a problem.

PR #480, opened 2022-04-15
implement join controlling functions

### LINKED ISSUE #427: Feature Request: join_many_to_one() and join_one_to_one()
I often need to perform join operations which confirm that the join is either 1:1 or many:1.  I have written functions for this in my personal package, but I think that they may be of general interest.  That said, they are a bit of a stretch for normal `janitor` functionality, so if you don't think it's a good fit, no worries.

billdenney:
Fixes #427

Implements join controlling functions.  See the examples in the help file for usage.


sfirke:
Would you like me to review this (a) in general (b) for inclusion in 2.2 this month?

billdenney:
Let's not jump on this yet.  The new dplyr functionality will need to integrate with this, so I'd rather wait for that than do something that may cause a conflict now.

sfirke:
Sounds good.  I am looking at the other requests/issues for v2.2 and sweating a little bit so it sounds like a good allocation of time for me to focus on those instead!

billdenney:
I think that the new dplyr functionality covers most of this PR, so I'm closing it now.

PR #483, opened 2022-05-15
clean_names() support for all types

### LINKED ISSUE #481: FR: support all classes by default in clean_names()
Hi,

This is a very minor FR, so feel free to reject it without any regard.

By default, `clean_names()` only support a few arbitrary classes and justifies it like this:

> `clean_names()` is intended to be used on data.frames and data.frame-like objects. For this reason there are methods to support using `clean_names()` on `sf` and `tbl_graph` (from tidygraph) objects. For cleaning other named objects like named lists and vectors, use `make_clean_names()`.

However, it seems that the dataframe method works out of the box for lists and vectors as well:

``` r
library(dplyr)
clean_names = janitor:::clean_names.data.frame 

c("A%"=1, "some naME"=2) %>% clean_names()
#>  a_percent some_na_me 
#>          1          2
list("A%"=1, "some naME"=2) %>% clean_names()
#> $a_percent
#> [1] 1
#> 
#> $some_na_me
#> [1] 2
```
<sup>Created on 2022-04-21 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

To tell, my code often ends up with this unattractive `:::` attribution.

Could you consider adding this behavior in `janitor:::clean_names.default()` instead?

DanChaltiel:
## Description
Enhance `clean_names.default()` to support all types of objects as long as they have either names or dimnames.

## Related Issue
fixes #481 

## Example
``` r
x = c("A%"=1, 2, 3)
names(x)
#> [1] "A%" ""   ""
janitor::clean_names(x)
#> a_percent         x       x_2 
#>         1         2         3

x = matrix(NA, nrow=2, ncol=2, dimnames=list(c("A B", "B C"), c("C D", "D E")))
janitor::clean_names(x)
#>     c_d d_e
#> a_b  NA  NA
#> b_c  NA  NA

x = 1:10
janitor::clean_names(x)
#> Error: `clean_names()` requires that either names or dimnames be non-null.
```

<sup>Created on 2022-05-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/R/clean_names.R
+++ b/R/clean_names.R
@@ -79,8 +79,7 @@ clean_names.default <- function(dat, ...) {
     return(dat)
   }
   if(is.null(names(dat))) {
-    colnames(dat) <- make_clean_names(colnames(dat), ...)
-    rownames(dat) <- make_clean_names(rownames(dat), ...)
+    dimnames(dat) <- lapply(dimnames(dat), make_clean_names, ...)
   } else {
     names(dat) <- make_clean_names(names(dat), ...)
   }
--- a/R/clean_names.R
+++ b/R/clean_names.R
@@ -76,7 +76,6 @@ clean_names.default <- function(dat, ...) {
       "`clean_names()` requires that either names or dimnames be non-null.",
       call. = FALSE
     )
-    return(dat)
   }
   if(is.null(names(dat))) {
     dimnames(dat) <- lapply(dimnames(dat), make_clean_names, ...)
--- a/R/clean_names.R
+++ b/R/clean_names.R
@@ -70,17 +70,21 @@ clean_names <- function(dat, ...) {
 
 #' @rdname clean_names
 #' @export
-clean_names.data.frame <- function(dat, ...) {
-  stats::setNames(dat, make_clean_names(names(dat), ...))
-}
-
-#' @rdname clean_names
-#' @export
 clean_names.default <- function(dat, ...) {
-  stop(
-    "No `clean_names()` method exists for the class ", paste(class(dat), collapse=", "),
-    "\nConsider janitor::make_clean_names() for other cases of manipulating vectors of names."
-  )
+  if(is.null(names(dat)) && is.null(dimnames(dat))) {
+    stop(
+      "`clean_names()` requires that either names or dimnames be non-null.",
+      call. = FALSE
+    )
+    return(dat)
+  }
+  if(is.null(names(dat))) {
+    colnames(dat) <- make_clean_names(colnames(dat), ...)
+    rownames(dat) <- make_clean_names(rownames(dat), ...)
+  } else {
+    names(dat) <- make_clean_names(names(dat), ...)
+  }
+  dat
 }
 
 #' @rdname clean_names
--- a/tests/testthat/test-clean-names.R
+++ b/tests/testthat/test-clean-names.R
@@ -392,6 +392,15 @@ test_that("Tests for clean_names.default() on lists and vectors", {
   )
 })
 
+test_that("Tests for clean_names.default() on arrays", {
+  x = array(NA, dim = c(2, 2, 2), dimnames = list(c("A", "B"), c("C", "D"), c("E", "F")))
+  clean = clean_names(x)
+  test = unlist(dimnames(x))
+  expect_false(isTRUE(all.equal(test, tolower(test))))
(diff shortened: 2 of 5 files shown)


billdenney:
Thanks for preparing this!  While it's an edge case, could you please extend it so that it works with arbitrarily many dimnames (e.g. a 3D array like `array(NA, dim = c(2, 2, 2), dimnames = list(row.names,
   column.names, matrix.names))

DanChaltiel:
Good point indeed!
I never use arrays but I agree this should be as generic as possible.

I re-used your case as a test, testing that the output is lowercase should be enough. My other tests were maybe a bit overkill, as I'm re-testing `make_clean_names()` features, I could remove them if needed.

billdenney:
With that one last change, I think we'd be good to merge.  @sfirke any other things you see?

tests/testthat/test-clean-names.R
```diff
@@ -371,14 +371,36 @@ test_that("Tests for cases beyond default snake", {
...
 })
 
-test_that("errors if not called on a data.frame", {
+test_that("Tests for clean_names.default() on lists and vectors", {
+  test_v <- seq_along(testing_vector)
+  names(test_v) <- testing_vector
+  test_list <- as.list(test_v)
+  
+  # Warnings due to partially handled mu
+  clean_v <- suppressWarnings(clean_names(test_v))
```
sfirke, line 380: Very minor thing but suppressWarnings on a test on this & following line makes me nervous.  It looks like the way @billdenney has addressed this is with `expect_warning()` instead, does it work to swap in that function?  Your comment in the preceding line helps.  Then if we go back, we know that the warning behavior is intentional and permanent.

R/clean_names.R
```diff
@@ -68,19 +68,22 @@ clean_names <- function(dat, ...) {
...
-  stop(
-    "No `clean_names()` method exists for the class ", paste(class(dat), collapse=", "),
-    "\nConsider janitor::make_clean_names() for other cases of manipulating vectors of names."
-  )
+  if(is.null(names(dat)) && is.null(dimnames(dat))) {
+    stop(
+      "`clean_names()` requires that either names or dimnames be non-null.",
+      call. = FALSE
+    )
+    return(dat)
```
sfirke, line 79: Could you add a test for this line?  Codecov is reporting a failure b/c this line is untested.  Though is it unreachable due to the preceding `stop()`?  If so it could just be removed.

sfirke (commented):
Looking good!  Thanks @DanChaltiel (and @billdenney for the review).  Two comments that I think are minor re: testing, if those are addressed I'm good to merge 👍

R/clean_names.R
DanChaltiel: The `return(dat)` was a leftover from the time this stop was a warning. I removed it

tests/testthat/test-clean-names.R
DanChaltiel: Indeed, expect_warning is much cleaner, I should have started with that.

DanChaltiel:
There you are, this should do 👍 

As a side comment, here is a quote from the book ["Advanced R"](https://adv-r.hadley.nz/conditions.html#custom-conditions):

> One of the challenges of error handling in R is that most functions generate one of the built-in conditions, which contain only a `message` and a `call`. That means that if you want to detect a specific type of error, you can only work with the text of the error message. This is error-prone, not only because the message might change over time, but also because messages can be translated into other languages.

Indeed, when you test your errors, you often rely on regexp. This means you cannot change the error/warning message as you might want. Since you already import rlang, you could swap every `stop("foobar")` to `rlang::abort("foobar", class="...")`, so that you can then call `expect_error(expr, class="...")` (and same for warnings).
I did that in my package crosstable a while ago and found this refactor was really worth it.

billdenney:
@DanChaltiel , I wish that all messages, warnings, and errors were classed, so that testing would be better, as you suggest.  I've started using classed messages, warnings, and errors some of the time in my code.  It is helpful.  If you add classes, please make the class name start with the package name so that there is a near-zero chance of conflicts.

sfirke:
Good idea on the classing of errors, Dan.  Thanks for this PR and for being receptive to tweaks!  I'm merging it now.

PR #358, opened 2020-04-05
Never duplicate names (Fix #251)

billdenney:
Fix #251

PR #497, opened 2022-12-01
Adds `allow_dupes`, fixes #495

### LINKED ISSUE #150: Add test for print.tabyl

### LINKED ISSUE #495: Feature request: allow users to opt-in for duplicate names in `make_clean_names()`
## Feature requests

There are times when users may wish to _opt-in_ to allow duplicate names when using `make_clean_names()`. One approach would be having an additional argument: `make_clean_names(..., allow_dupes = FALSE, ...)`. Here is [some support](https://stackoverflow.com/q/62594765/2572423).

I believe a solution would be bypassing the final `while` loop in the `make_clean_names()` source (which was introduced in https://github.com/sfirke/janitor/pull/358):

```
...
cased_names <- snakecase::to_any_case(made_names, case = case, 
    sep_in = sep_in, transliterations = transliterations, 
    parsing_option = parsing_option, numerals = numerals, 
    ...)
while (any(duplicated(cased_names))) {
    dupe_count <- vapply(seq_along(cased_names), function(i) {
        sum(cased_names[i] == cased_names[1:i])
    }, 1L)
    cased_names[dupe_count > 1] <- paste(cased_names[dupe_count > 
        1], dupe_count[dupe_count > 1], sep = "_")
}
cased_names
```

So effectively:

```
...
cased_names <- snakecase::to_any_case(made_names, case = case, 
    sep_in = sep_in, transliterations = transliterations, 
    parsing_option = parsing_option, numerals = numerals, 
    ...)

if (!allow_dupes) {
  while (any(duplicated(cased_names))) {
    dupe_count <- vapply(seq_along(cased_names), function(i) {
      sum(cased_names[i] == cased_names[1:i])
    }, 1L)
    cased_names[dupe_count > 1] <- paste(cased_names[dupe_count > 
                                                       1], dupe_count[dupe_count > 1], sep = "_")
  }
}
cased_names
```

JasonAizkalns:
## Description

Adds an `allow_dupes` parameter to `make_clean_names()` -- defaults to `FALSE`.

## Related Issue

Fixes #495

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,7 +8,7 @@
 
 * A minor breaking change is that the time zone is now always set for `excel_numeric_to_date()` and `convert_date()`.  The default timezone is `Sys.timezone()`, previously it was an empty string (`""`). (#422, thanks **@billdenney** for fixing)
 
-* A minor breaking change effects `make_clean_names()` (and therefore `clean_names()`). `make_clean_names()` now uses the `unique_sep` argument from `snakecase::make_any_case()` to handle de-duplication of names. The incremental suffix counter is now one less than in the past (i.e., a de-duplicated variable with a suffix of `_2` becomes `_1`, `_3` becomes `_2`, etc.). This change also results in a new feature and the ability to allow for duplicate names by setting `unique_sep = NULL`. (#495, thanks **@JasonAizkalns** for fixing and **@billdenney** and **@sfrike** for the guidance)
+* A minor breaking change affects `make_clean_names()` (and therefore `clean_names()`). `make_clean_names()` now uses the `unique_sep` argument from `snakecase::to_any_case()` to handle de-duplication of names. The incremental suffix counter is now one less than in the past (i.e., a de-duplicated variable with a suffix of `_2` becomes `_1`, `_3` becomes `_2`, etc.). This change also results in a new feature and the ability to allow for duplicate names by setting `unique_sep = NULL`. (#495, thanks **@JasonAizkalns** for fixing and **@billdenney** and **@sfrike** for the guidance)
 
 ## New features
 
@@ -29,8 +29,7 @@
 
 * `clean_names()` now supports all object types that have either names or dimnames (#481, @DanChaltiel).
 
-* `make_clean_names()` now allows duplicate names to be returned with the `allow_dupes = TRUE`
-argument (#495, @JasonAizkalns).
+* `make_clean_names()` allows duplicate names to be returned by specifying `unique_sep = NULL` (#495, @JasonAizkalns).
 
 ## Bug fixes
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@
 
 * A minor breaking change is that the time zone is now always set for `excel_numeric_to_date()` and `convert_date()`.  The default timezone is `Sys.timezone()`, previously it was an empty string (`""`). (#422, thanks **@billdenney** for fixing)
 
+* A minor breaking change effects `make_clean_names()` (and therefore `clean_names()`). `make_clean_names()` now uses the `unique_sep` argument from `snakecase::make_any_case()` to handle de-duplication of names. The incremental suffix counter is now one less than in the past (i.e., a de-duplicated variable with a suffix of `_2` becomes `_1`, `_3` becomes `_2`, etc.). This change also results in a new feature and the ability to allow for duplicate names by setting `unique_sep = NULL`. (#495, thanks **@JasonAizkalns** for fixing and **@billdenney** and **@sfrike** for the guidance)
+
 ## New features
 
 * `row_to_names()` now has a new helper function, `find_header()` to help find the row that contains the names.  It can be used by passing `row_number="find_header"`, and see the documentation of `row_to_names()` and `find_header()` for more examples. (fix #429)
--- a/NEWS.md
+++ b/NEWS.md
@@ -27,6 +27,9 @@
 
 * `clean_names()` now supports all object types that have either names or dimnames (#481, @DanChaltiel).
 
+* `make_clean_names()` now allows duplicate names to be returned with the `allow_dupes = TRUE`
+argument (#495, @JasonAizkalns).
+
 ## Bug fixes
 
 * `adorn_percentages()` was refactored for compatibility with `dplyr` package versions > 1.0.99 (#490)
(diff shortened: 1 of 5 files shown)


billdenney:
With a first look, the PR looks pretty good to me.  Thanks!

One question and one thought:

* Did you rebuild the documentation with `devtools::document()` or similar?  There is a change in the .Rd files that looks like it may have been manually made (but, I'm not sure of that).
* Should we specifically exclude this option from `clean_names()` since it would result in invalid data.frame/tibble names?

JasonAizkalns:
Yes, I rebuilt the documentation via `devtools::document()`. Simply added the param, typed it up and decorated it in `make_clean_names.R` then ran `devtools::document()`. It triggered changes to `clean_names.Rd` and `make_clean_names.Rd` which I committed. It also triggered changes to `convert_to_date.Rd`, `describe_class.Rd` and `DESCRIPTION` but I did not commit those.

Yes, I like the idea and agree with excluding this option from `clean_names()` -- what's the best way to go about doing that?

I had an additional thought around the documentation. I'm not thrilled with the inherited dot parameters from `snakecase::to_any_case` (`@inheritDotParams snakecase::to_any_case`) -- specifically, because of the `unique_sep` argument. Per the inherited docs, it states:

> unique_sep
A string. If not NULL, then duplicated names will get a suffix integer in the order of their appearance. The suffix is separated by the supplied string to this argument.

...but if you set `unique_sep = NULL` explicitly, the "while loop" still runs. So a bit misleading. Perhaps a new/separate issue?

billdenney:
To check for the `allow_dupes` in the arguments to `clean_names()`, I would do something like

```r
if ("allow_dupes" %in% names(list(...)) {
  stop("`allow_dupes` is not an allowed argument for `clean_names()` as it can lead to invalid names")
}
```

or something like that.

But your other point is more important.  Are we doing too much work inside `make_clean_names()`?  Could we have the deduplication of names by setting `unique_sep="_"`?  The (big) problem I see with that is it would be a breaking change since doing that would change the current output.

JasonAizkalns:
Yea, not sure. It seems like if you made the default of both `clean_names()` and `make_clean_names()` to set `unique_sep="_"` that you effectively achieve the same result of the "while loop" code but the indexing of duplicates changes.

The incremental counter does change -- it appends `_1`, `_2`, `_3`, ... whereas the while loop will append starting at `_2`:

```r
# Removed while loop code:
make_clean_names(c("a", "a 2", "a 2", "a_2"), unique_sep = "_")
[1] "a"     "a_2"   "a_2_1" "a_2_2"
# Which means that doing this:
make_clean_names(c("a", "a 2", "a 2", "a_2"))
[1] "a"   "a_2" "a_2" "a_2"
# would be the same as explicitly setting unique_sep = NULL (default)
# make_clean_names(c("a", "a 2", "a 2", "a_2"), unique_sep = NULL)

# As-is, with while loop:
make_clean_names(c("a", "a 2", "a 2", "a_2"))
[1] "a"     "a_2"   "a_2_2" "a_2_3"
# which is the same as... make_clean_names(c("a", "a 2", "a 2", "a_2"), unique_sep = NULL)
```

The more I think about this. I don't think I like the `allow_dupes` option any more. I'm actually thinking that the while loop introduced in #358 created some friction with `unique_sep`. 

What do you think?

billdenney:
I'm going to move the conversation back to the issue for the moment.

tests/testthat/test-clean-names.R
```diff
@@ -184,7 +184,21 @@ test_that("locale-specific make_clean_names tests", {
 test_that("do not create duplicates (fix #251)", {
   expect_equal(
     make_clean_names(c("a", "a", "a_2")),
```
sfirke, line 186: Nit: this `a_2` was to test a collision, it should now be `a_1` to collide with the new result on the second `a`.  I believe the result on the next line would be `c("a", "a_2", "a_1")`

NEWS.md
```diff
@@ -27,6 +29,9 @@
 
 * `clean_names()` now supports all object types that have either names or dimnames (#481, @DanChaltiel).
 
+* `make_clean_names()` now allows duplicate names to be returned with the `allow_dupes = TRUE`
+argument (#495, @JasonAizkalns).
```
sfirke, line 33: Should this now be:

> allows duplicate names to be returned by specifying `unique_sep = NULL`

NEWS.md
```diff
@@ -8,6 +8,8 @@
 
 * A minor breaking change is that the time zone is now always set for `excel_numeric_to_date()` and `convert_date()`.  The default timezone is `Sys.timezone()`, previously it was an empty string (`""`). (#422, thanks **@billdenney** for fixing)
 
+* A minor breaking change effects `make_clean_names()` (and therefore `clean_names()`). `make_clean_names()` now uses the `unique_sep` argument from `snakecase::make_any_case()` to handle de-duplication of names. The incremental suffix counter is now one less than in the past (i.e., a de-duplicated variable with a suffix of `_2` becomes `_1`, `_3` becomes `_2`, etc.). This change also results in a new feature and the ability to allow for duplicate names by setting `unique_sep = NULL`. (#495, thanks **@JasonAizkalns** for fixing and **@billdenney** and **@sfrike** for the guidance)
```
sfirke, line 11: Nits: this should be `affects` instead of `effects` and `snakecase::to_any_case`

sfirke (changes requested):
Gotta love adding a feature by shortening the source code 🙌   I have three little nits to fix, two in the docs and one little test.  Then this is good to merge!

sfirke:
ping @JasonAizkalns - do you think you'll be able to make those edits I suggested?  They should be trivial, I estimate 15 minutes or less.  Then we can get this merged.

Looks like we're heading toward a CRAN release soon to beat some changes to dplyr, this is one PR we can square away before then.  Thanks again!

tests/testthat/test-clean-names.R
JasonAizkalns: Other two resolved, but this one I'm not sure. I still believe this should result in `c("a", "a_1", "a_2")` -- the only duplicates are the two versions of "a". Should we add an additional test for `c("a", "a", "a_1")` --> `c("a", "a_2", "a_1")` or change _this_ test?

JasonAizkalns: I went ahead and added another test to force this new type of collision. Kept the other one for protection of a "normal" no dupes case.

sfirke: Perfect.

sfirke:
Merged.  Thanks @JasonAizkalns for the excellent PR! 🙏

PR #502, opened 2022-12-12
Add single_value()

### LINKED ISSUE #428: Feature Request: single_value()
This is a function that would consider some values to be missing, but for all non-missing values, it would ensure that they have the same value.

I often work with datasets where I need to combine information for subjects in clinical trials.  For that, I need to ensure that I have the same information from each of the different sources.  For example, I may have multiple sources for the age of a subject when they start the study.

When I combine those data sets, I need to end up with the age as the same across all data.  A paradigm I often use is below.  Would that be of interest?

``` r
library(tidyverse)
library(bsd.report)

my_data_good <-
  tibble(
    Subject=rep(1:2, each=2),
    Age=c(1, NA, 2, NA)
  ) %>%
  group_by(Subject) %>%
  mutate(
    Age=single_value(Age)
  )
my_data_good
#> # A tibble: 4 x 2
#> # Groups:   Subject [2]
#>   Subject   Age
#>     <int> <dbl>
#> 1       1     1
#> 2       1     1
#> 3       2     2
#> 4       2     2

my_data_bad <-
  tibble(
    Subject=rep(1:2, each=2),
    Age=c(1, NA, 2, 3)
  ) %>%
  group_by(Subject) %>%
  mutate(
    Age=single_value(Age)
  )
#> Error: Problem with `mutate()` input `Age`.
#> x More than one (2) value found (2, 3)
#> i Input `Age` is `single_value(Age)`.
#> i The error occurred in group 2: Subject = 2.
```

<sup>Created on 2021-02-04 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

billdenney:
Fix #428 

This only handles the vector case as I don't see a clean way to handle the checker/assert function case at the same time.


billdenney:
@sfirke, ping on this one.

tests/testthat/test-single_value.R
```diff
@@ -0,0 +1,17 @@
+test_that("single_value", {
+  expect_equal(single_value(1), 1)
+  expect_equal(single_value(c(1, NA)), 1)
+  expect_equal(single_value(c(1, NA, 1)), 1)
+  expect_equal(single_value(NA_real_), NA)
+  expect_equal(single_value(NA, missing=NA_real_), NA_real_)
```
sfirke, line 6: I'm unclear on what this is testing.

sfirke, line 6: Want to add a test for multiple values of missing where the first takes precedence?  E.g., `single_value(c(""), missing = c(NA, ""))` correctly returns NA.

R/single_value.R
```diff
@@ -0,0 +1,29 @@
...
+#' @param missing The vector of values to consider missing in \code{x}
+#' @param warn_if_all_missing Generate a warning if all values are missing?
+#' @param info If more than one value is found, append this to the warning or
+#'   error to assist with determining the location of the issue.
+#' @return \code{x} as the scalar single value found throughout (or an error if
+#'   more than one value is found).
+#' @examples
+#' single_value(c(NA, 1))
+#' @export
+single_value <- function(x, missing=NA, warn_if_all_missing=FALSE, info=NULL) {
```
sfirke, line 16: I don't see `warn_if_all_missing` used in the function, e.g., `single_value(c(""), missing = c(NA, ""), warn_if_all_missing = TRUE)` does not warn.  But I think it's a useful feature to implement.

sfirke, line 16: This might be odd style but what do you think about `missing=c(NA)` as the default argument?  Then it hints to the user that a custom `missing` can take multiple values.

R/single_value.R
```diff
@@ -0,0 +1,29 @@
+#' Ensure that a vector has only a single value throughout.
+#'
+#' Missing values are replaced with the single value, and if all values are
+#' missing, the first value in \code{missing} is used throughout.
+#'
+#' @param x The vector which should have a single value
+#' @param missing The vector of values to consider missing in \code{x}
+#' @param warn_if_all_missing Generate a warning if all values are missing?
+#' @param info If more than one value is found, append this to the warning or
+#'   error to assist with determining the location of the issue.
```
sfirke, line 10: Can you give an example of how to use `info`?  I see how it works but not when it would be handy.  E.g., if you're using this with `group_by` can it tell you which grouping var has the multiple variables?

R/single_value.R
```diff
@@ -0,0 +1,29 @@
...
+#' @return \code{x} as the scalar single value found throughout (or an error if
+#'   more than one value is found).
+#' @examples
+#' single_value(c(NA, 1))
```
sfirke, line 14: What do you think about this for `@examples` to show a typical use case?  Like what you showed in the original issue for this function.
```
library(dplyr)
# imputes a value where missing:
dat %>%
  dplyr::group_by(id) %>%
  mutate(age = single_value(age))

# fails because of multiple values:
\dontrun{
dat %>%
  dplyr::group_by(id) %>%
  mutate(country = single_value(country))
}
```

sfirke (commented):
I took a look, my comments and suggestions are in the review.  This will be a nice addition!

tests/testthat/test-single_value.R
billdenney: Good call.  Added.

R/single_value.R
billdenney: I'll add that as an example, instead.  That will hopefully make it clearer (and I try to avoid the odd style... 😄 ).

billdenney: Whoops!  Fixed and tested.

R/single_value.R
billdenney: I added an example.

R/single_value.R
billdenney: I added a similar example.

billdenney:
@sfirke, I think that I've addressed everything.

R/single_value.R
```diff
@@ -11,10 +11,37 @@
...
+#'   )
+#'
+#' try(
+#' # info is useful to give when multiple values may be found to see what
+#' # grouping variable or what calculation is causing the error
+#' data.frame(A = rep(1:3, each = 2),
+#'            B = c(rep(1:2, each = 2), 1:2)) %>%
+#'   dplyr::group_by(A) %>%
+#'   dplyr::mutate(
+#'     C = single_value(B, info = paste("Calculating C for group A=", A))
```
sfirke, line 36: ah this is good!  I hadn't grokked that one would `paste()` the grouping variable in there so didn't see the utility.  Great that you added it as an example.

sfirke (approved):
LGTM!

R/single_value.R
billdenney: The other utility is indicating the column name being calculated as that isn't always clear in the dplyr error messages.  When I use that option, I use it for both finding the problem group and the problem column.

PR #504, opened 2022-12-12
Switch to using testthat edition 3

billdenney:
This looks like a whole lot of changes, but the only real goals were to do three things:

* Switch to testthat edition 3,
* make testing (mostly) silent when successful, and
* update to current GitHub continuous integration files

There are no functionality updates here.


sfirke (approved):
This PR warms my heart!  I would not have known how to update the Github actions and the refactoring of tests is an impressive feat.  Fully reviewed, LGTM.

sfirke:
Looks like CI failed due on the tests for `clean_names` where we changed numbering of duplicates in #497 .  I just re-triggered the runs in case that fixes it, I see you merged main branch back to this PR.

sfirke:
Looks like two tests are still failing.

billdenney:
I fixed the failing tests.  While I was in there, I noticed that `adorn_totals()` was being tested in two separate files, so I combined them.  If there was a reason for two files, we can revert that last commit.

sfirke:
>  If there was a reason for two files, we can revert that last commit.

Nice catch!  Looks like one of those files was added in #413 and only ever edited once, so no, it was an error.  Thanks for re-combining.

PR #505, opened 2022-12-12
Add get_one_to_one()

### LINKED ISSUE #291: Feature Request: Detect Columns that Map to Each Other
I often receive data with columns that contain either identical or effectively identical columns.

As an example, I may have a column indicating a human-readable medical laboratory test (like "Cholesterol, LDL"), and another column indicating a more computer-friendly standardized test number (called a LOINC code that may be 12345).  It would be helpful to me to have a reporter function that indicates the columns that all map to the same values.  The interface could look like:

`find_mapping_cols(x)`

where it would return a list with elements that are character vectors of columns that map to each other:

```r
foo <- data.frame(
  Lab_Test_Long=c("Cholesterol, LDL", "Cholesterol, LDL", "Glucose"),
  Lab_Test_Short=c("CLDL", "CLDL", "GLUC"),
  LOINC=c(12345, 12345, 54321),
  Person=c("Sam", "Bill", "Sam"),
  stringsAsFactors=FALSE
)
find_mapping_cols(foo)
```

That would return a list that looks like:

```r
list(c("Lab_Test_Long", "Lab_Test_Short", "LOINC"), "Person")
```

It's up to the user what they would want to do with that list (because which column of the set is most useful to keep is not obvious).

billdenney:
Fix #291


billdenney:
@sfirke, This should be ready for review, now.  I removed the need for the `forcats` package.

sfirke:
When I run `get_one_to_one(mtcars %>% slice(1:3))` the list items 1, 6, and 9 are great.  What's the value of having `cyl` listed by itself in item 2 when it just appeared with mpg in item 1?  I'd be expecting each variable name to appear only once in the output.

billdenney:
Happy New Year!  And, thanks for finding that issue.  I was mis-using a for loop where I needed a while loop.  It should be fixed now.

sfirke:
Yep that's fixed, thanks.

Now that I've tried this out, I'm more excited to use it!  It's handy.  I also have a further thought, sorry to delabor this PR.

When I do `get_one_to_one(mtcars[1:5, ])` I don't find it useful to see list entries 3-5 before the 6th, `"am" "gear"`.  I only want positive hits for one-to-one.  And if this were a larger data.frame, I could see it getting tedious to scroll through all the entries looking for ones where there are multiple variables.

I like this result better:
```
x <- get_one_to_one(mtcars[1:5, ])
x[lapply(x, length) > 1]

[[1]]
[1] "mpg"  "disp" "drat"

[[2]]
[1] "cyl" "hp" 

[[3]]
[1] "am"   "gear"

```

This changed behavior even matches the function description right now.  What do you think?

And then if that change was made, I'd suggest that when there are no one-to-one columns, like with `get_one_to_one(mtcars)`, it return a list of length 0 or whatever the natural output is along with a message that no columns in data.frame {input name} are one-to-one.

sfirke:
Also, I'll add it to a vignette before the release.  I'll open a new issue for that now.

billdenney:
Good call.  I just updated it to drop non-mapping columns and added a message.  (And, thanks for vignetting it.)

sfirke:
Looks good to me 🙌   Nice work and happy new year to you too!

PR #507, opened 2022-12-28
clean_names for dbplyr database connections

sfirke:
Don't review yet.  I'm starting the PR to trigger CI and see what happens on Github.


sfirke:
I don't see how to set up the db to test against.  I posted on StackOverflow, today might not be a good day to get an answer.  Also pinged the dbplyr team from issue 752.  I'll try again next week, if nothing else maybe I can replicate dbplyr tests: https://github.com/tidyverse/dbplyr/blob/main/tests/testthat/test-backend-mysql.R

billdenney:
It looks like you may be able to create a SQLite temporary database for testing (that would add RSQLite as another Suggests package in the DESCRIPTION file):

https://dbplyr.tidyverse.org/reference/testing.html#ref-examples

mgirlich:
@sfirke I saw your comment in the dbplyr issue and I think I know the issue. It works locally because you have RSQLite installed but it fails in the CI because it doesn't install RSQLite. This is because there is no direct link between janitor and RSQLite. You could:

* add RSQLite as a Suggested package (I think this is quite typical for packages needed for testing)
* modify the Github action, see [Setup R dependencies](https://github.com/r-lib/actions/tree/v2/setup-r-dependencies#readme)

sfirke:
Thanks @mgirlich, adding RSQLite to `Suggests` did it!  @billdenney this is finished, would you be up for reviewing when you get a chance?

R/clean_names.R
```diff
@@ -119,6 +120,20 @@ clean_names.tbl_graph <- function(dat, ...) {
   dplyr::rename_all(dat, .funs=make_clean_names, ...)
 }
 
+#' @rdname clean_names
+#' @export
+#' @importFrom dplyr rename_with
```
billdenney, line 125: Since you're using `dplyr::rename_with` below (referencing the package::), you shouldn't need this `@importFrom dplyr rename_with`

R/clean_names.R
```diff
@@ -119,6 +120,20 @@ clean_names.tbl_graph <- function(dat, ...) {
   dplyr::rename_all(dat, .funs=make_clean_names, ...)
```
billdenney, line 120: Unrelated, but can you change this to `rename_with` instead of `rename_all` since `rename_with` is the currently-preferred version?

billdenney (commented):
Looks good overall.  One minor, style comment and another unrelated one that should prevent future headaches.

R/clean_names.R
sfirke: I tried changing this line to `dplyr::rename_with(dat, janitor::make_clean_names, .cols = dplyr::everything(), ...)` and I got a failed test:
```
-- Error (test-clean-names.R:596:3): tbl_graph/tidygraph -----------------------
Error: no applicable method for 'rename_with' applied to an object of class "c('tbl_graph', 'igraph')"
```
I like the idea of updating this now-superseded function.  Do you see a reasonable way forward or should we just punt on this one for now?

R/clean_names.R
sfirke: Nice catch, I took it out.

sfirke: Also removed it from the tidygraph section too.

R/clean_names.R
billdenney: I would leave it alone for now.  I don't see a clean way to fix it.

sfirke:
Okay then I think we're set. Want to merge it if you think it's ready?

billdenney:
Alacazam!

PR #511, opened 2023-01-12
get_dupes sorts first on desc(dupe_count)

### LINKED ISSUE #493: Suggestion: get_dupes results should be sorted by dupe_count descending
## Feature requests

Sort the results of `get_dupes` by `dupe_count`.

## Remarks

Right now it sorts alphabetically by the grouping variable.  But usually I then add `%>% arrange(desc(dupe_count))` to start with the most-duplicated combination.  I suppose this would potentially be a breaking change for some use cases.  But I think it is ultimately more useful than the alphabetical sort.

## Example
`mtcars %>% get_dupes(cyl)`

In my result, the group with `cyl = 8` that has 14 records would appear first.

This would be easy to implement.  Thoughts from others?

sfirke:
fixes #493


sfirke:
Extremely straightforward code modification so merged w/o review

Issue #513, opened 2023-01-16
Changes to make_clean_names() breaks packages on CRAN


sfirke:
Re: this matter https://github.com/sfirke/janitor/issues/495#issuecomment-1335377407

The breaking change to how duplicate names are assigned suffixes `_1` `_2` etc. breaks tests in at least the BFS and nflfastR packages.
To resolve it, I see at least these options:

1. Refactor that feature so that it adds an allow_dupes argument and preserves the existing behavior of `make_clean_names`.  Submit this month to CRAN.
2. Same as Option 1 but do it after the 2.2 release, pulling this change out for now.
3. Pull the change out for this 2.2 release, then merge back into main as-is with the breaking change.  Notify the affected package maintainers, help them fix if necessary, give them time to resubmit to CRAN, and then submit janitor to CRAN again.

I lean toward 1 (if we have bandwidth now) or 2 (if not).  Seeing a couple of CRAN packages that depended on the old behavior makes me think of all the users whose scripts will be similarly affected.  I don't think our change in the `make_clean_names` output is an improvement, it's neutral IMO - and while it improves the janitor codebase, I think avoiding major disruption outweighs that.

sfirke:
Sounds like either way the immediate move is to remove the breaking change for 2.2 to buy some time.  Time for me to try out `git revert`.

I wonder also if the next release would be the a good time to implement #240 and split the package up, especially if janitor became tidytab and custodian (??) or something and janitor got left behind in maintenance mode...

billdenney:
I like the idea of implementing #240 (as I have for a while).  I'd lean toward janitor keeping the name and `tidytab` being split off.  (I like the name janitor, and it is well-known in the R community, too!)  The breaking change of `make_clean_names()` giving slightly different numbering doesn't seem like it's worth a whole-scale switch in the package, to me.

billdenney:
I've been thinking a bit more about it.  How about we do the following for the release:

* Make a change for `make_clean_names(janitor_version = "2.1.0")` which will use the old behavior with a deprecation warning.
* If the `janitor_version` option is NULL or missing, it uses the new method.
* Suggest PRs for the new behavior for the dependent packages.
* Switch the default to `janitor_version = NULL` in the next version of `janitor` (after the other packages have released on CRAN-- and probably in at least 6 months)
* Remove the `janitor_version = "2.1.0"` code some time in the relatively distant future (at least 2024)

sfirke:
[...]

PR #519, opened 2023-01-29
control allowing of duplicate names with allow_dupes

sfirke:
Implements #495 using initial version of #497 to address issues identified and solution proposed in #513


billdenney:
Looks good to me.  Merged.

PR #522, opened 2023-02-01
show new functions in catalog vignette

### LINKED ISSUE #508: Create vignette entries for new functions
Add to the main vignette.  Use the current example data.frame if possible, otherwise maybe a subset of starwars.

- [x] #505 
- [x] #502 
- [ ] #443 (see #453)

sfirke:
fixes #508


sfirke:
Didn't update the .md vignette (a) because the file generated in my local environment had some weird transliteration issues in the output (b) I am rusty and don't totally remember the role of the .md output anyway?

PR #538, opened 2023-04-18
Add paste_skip_na()

### LINKED ISSUE #537: Feature Request: `paste_skip_na()` function that skips NA values when pasting
I often want to paste multiple values together skipping `NA` values.

This could be related to #536 as it could be used for combining header values there.

This is how paste works:

```r
paste("A", NA, sep = "_")
# [1] "A_NA"
paste(NA, NA, sep = "_")
# [1] "NA_NA"
```

The proposed interface would be:

```r
paste_skip_na("A", NA, sep = "_")
# [1] "A"
paste_skip_na(NA, NA, sep = "_")
# [1] NA_character_
```

The `collapse` and `recycle0` arguments of `paste()` should also be handled consistently with `paste()`.  `collapse` would `paste_skip_na()` again, and I'd have to look into how the `recycle0` argument works now.

I have a function that does most of this, if interested.

billdenney:
Fix #537


billdenney:
@sfirke, this is ready for your review.

R/paste_skip_na.R
```diff
@@ -0,0 +1,64 @@
...
+    } else if (!is.null(collapse)) {
+      if (all(is.na(args[[1]]))) {
+        # Collapsing with all NA values results in NA
+        NA_character_
+      } else {
+        # Collapsing without all NA values collapses the non-NA values
+        paste(na.omit(args[[1]]), sep=sep, collapse=collapse)
+      }
+    } else {
+      # as.character() to ensure that logical NA values are converted to
```
sfirke, line 28: `paste_skip_na(c("A", "B", "C", NA))` returns the input.  That is better than `paste()` which quotes the `NA` -- but as a naive user I assumed this would return `c("A", "B", "C")`.  I think I'm missing something but what?

R/paste_skip_na.R
```diff
@@ -0,0 +1,64 @@
...
+#' paste_skip_na(NA) # NA_character_
+#' paste_skip_na("A", NA) # "A"
+#' paste_skip_na("A", NA, c(NA, "B"), sep = ",") # c("A", "A,B")
+#' @export
+paste_skip_na <- function(..., sep=" ", collapse=NULL) {
+  args <- list(...)
+  if (length(args) <= 1) {
+    if (length(args) == 0) {
+      # match the behavior of paste
+      paste(sep=sep, collapse=collapse)
```
sfirke, line 18: Nit: could you add spaces like `sep = sep` as you have in the example on line 11?

R/paste_skip_na.R
```diff
@@ -0,0 +1,64 @@
...
+        # Collapsing without all NA values collapses the non-NA values
+        paste(na.omit(args[[1]]), sep=sep, collapse=collapse)
+      }
+    } else {
+      # as.character() to ensure that logical NA values are converted to
+      # NA_character_
+      as.character(args[[1]])
+    }
+  } else {
+    # There are at least 2 arguments; paste the first two and recurse
```
sfirke, line 33: This recursion is elegant!

R/paste_skip_na.R
```diff
@@ -0,0 +1,64 @@
+#' Concatenate strings dropping missing values
```
sfirke, line 1: What do you think about:

 > Like \code{paste()} but missing values are omitted.

Since the docs point mostly to `paste()` already.

tests/testthat/test-paste_skip_na.R
```diff
@@ -0,0 +1,25 @@
+test_that("paste_skip_na", {
+  # handle no arguments the same as paste()
+  expect_equal(paste_skip_na(), paste())
+  expect_equal(paste_skip_na(NA), NA_character_)
+  expect_equal(paste_skip_na(NA, NA), NA_character_)
+  expect_equal(paste_skip_na(NA, NA, sep = ","), NA_character_)
+  # It does not behave like paste(NA, NA, collapse = ",") nor does it behave like paste(c(), collapse = ",")
```
sfirke, line 7: huge improvement that these are both simply `NA` - that's data janitor-ing!

sfirke (commented):
Nicely done!  I just have a question and a few small nits.

R/paste_skip_na.R
billdenney: I have a strong preference that the input length is the same as the output length.  That way, the output can be used as part of a tidy pipeline.  The following code requires the same length input and output:

```
data.frame(A = c("B", NA)) %>% mutate(C = paste(A))
```

sfirke: Ah of course, I was only thinking about single values and not vectorization on a column.  100% agree.

R/paste_skip_na.R
billdenney: I intend to change my coding style to that, but it's still not in my muscle memory.  😄

R/paste_skip_na.R
billdenney: That's better!

sfirke:
Merged. Thanks, Bill!

PR #542, opened 2023-05-17
Allow multiple rows input to row_to_names()

### LINKED ISSUE #536: Feature suggestion: allow multiple rows input to `row_to_names()`
Hi there!
I made for my personal use two functions that I believe might be of interest for janitor.
The use case is an Excel table containing variable names in multiple rows, while also having merged cells in some of them.
This code features two functions: `merge_names()` and `fix_names()` (which is currently serving only as a utility function for `merge_names()`, but can also be considered for standalone use).

```
library(dplyr)
library(tidyr)
library(stringr)
library(purrr)
```

`fix_names()` helps to fix column names of Excel tables. For an example in the wild. see those found in [Israeli CBS municipality data](https://www.cbs.gov.il/he/publications/doclib/2019/hamakomiot1999_2017/2018.xlsx) (the link would download an Excel file; Hebrew alert).
It does so by fixing a single row of to-be column names.
It receives a data frame, an integer number representing a single row considered as holding (some of the) variable names, and a logical length-one vector specifying whether to fill missing values with preceding values or not.
This last argument is mostly used in the case of merged cells in Excel files.
The function first transposes the row, and then either fills it with values or turns NAs to empty strings.

```
fix_names <- function(data, row_num, fill_missing) {
  data <- data |> 
    slice(row_num) |> 
    pivot_longer(everything()) |> 
    select(value)
  
  if (fill_missing) {
    data <- data |> 
      fill(value)
  }
  else {
    data <- data |> 
      replace_na(list(value = "")) |> 
      mutate(
        value = if_else(
          str_length(value) > 0,
          str_c(" ", value),
          value
        )
      )
  }
  data |> 
    pull(value)
}
```

This function fixes column names of Excel tables. For an example in the wild. see those found in [Israeli CBS municipality data](https://www.cbs.gov.il/he/publications/doclib/2019/hamakomiot1999_2017/2018.xlsx) (the link would download an Excel file; Hebrew alert).
It receives a data frame, a vector with the row numbers considered as holding the variable names, and a logical vector specifying whether to fill missing values with preceding values or not.
This last argument should be either of length 1, or of length of `row_num`.
It is mostly used in the case of merged cells in Excel files.
The function saves the new column names by iterating over every row, fixing the variable names, binding all names to a single tibble, uniting the different columns to a single column and pulling these values as a vector. The old rows containg these column names are filtered out, and the new merged and fixed names are set to the data frame, which is then returned.

```
merge_names <- function(data, names_num = 2, fill_missing = TRUE) {
  rows <- seq(names_num)
  col_names_merged <- map2(
    rows,
    fill_missing,
    \(i, fill_missing) fix_names(data, row_num = i, fill_missing)
  ) |> 
    bind_cols() |> 
    unite(col = "var_names", sep = "") |> 
    pull(var_names)

  data <- data |>
    slice(-rows)
  
  names(data) <- col_names_merged
  
  data
}
```

Let me know if you think this is of interest, and if so, I can create a PR.
Thanks!

matanhakim:
## Description
This is mostly useful when it comes to data from Excel spreadsheets, where variable names are spread over multiple rows.
Please note that this is not yet a totally finished PR, as one test still fails and no new tests were built yet.
I would like to receive some feedback regarding this to understand if this is even the right direction, since this is my first PR contribution to an established package.

Changes are mostly around allowing multiple rows input and handling NA's and missing values.
Additionally, a test was removed and some code style changes were made for better readability.

## Related Issue
Fixes #536.


R/row_to_names.R
```diff
@@ -22,16 +22,14 @@
 #' x %>%
 #'   row_to_names(row_number = "find_header")
 #' @export
-row_to_names <- function(dat, row_number, ..., remove_row = TRUE, remove_rows_above = TRUE) {
+row_to_names <- function(dat, row_number, ..., remove_row = TRUE, remove_rows_above = TRUE, sep = "_") {
```
billdenney, line 25: The `sep` argument will need documentation.

R/row_to_names.R
```diff
@@ -22,16 +22,14 @@
...
   # Check inputs
   if (!(is.logical(remove_row) & length(remove_row) == 1)) {
     stop("remove_row must be either TRUE or FALSE, not ", as.character(remove_row))
   } else if (!(is.logical(remove_rows_above) & length(remove_rows_above) == 1)) {
     stop("remove_rows_above must be either TRUE or FALSE, not ", as.character(remove_rows_above))
-  } else if (length(row_number) != 1) {
-    stop("row_number must be a scalar")
   }
-  if (row_number %in% "find_header") {
+  if (any(row_number %in% "find_header")) {
```
billdenney, line 32: This would be better as `identical(row_number, "find_header")` because having some rows as numbers and one as `"find_header"` would not be an input we would want to support.

R/row_to_names.R
```diff
@@ -44,7 +42,16 @@ row_to_names <- function(dat, row_number, ..., remove_row = TRUE, remove_rows_ab
...
-  new_names <- as.character(unlist(dat[row_number, ], use.names = FALSE))
+  new_names <- dat[row_number, ] %>% 
+    lapply(reduce_na) %>% 
+    as.data.frame() %>% 
+    replace(is.na(.), "") %>% 
+    lapply(paste_skip_na, collapse = sep) %>% 
+    lapply(stringr::str_replace_all, pattern = "__+", replacement = "_") %>% 
+    lapply(stringr::str_replace_all, pattern = "_$", replacement = "") %>% 
+    unlist(use.names = FALSE) %>% 
+    as.character()
```
billdenney, line 53: I think that all of this could be simplified to `sapply(dat[row_number, ], paste_skip_na, collapse = sep)`, right?  (Maybe I'm missing something.)  The downside is that when all values are `NA`, it would have a column name that is actually `NA` rather than the character string `"NA"`.  The other downside is that you'll have sequential underscores and underscores at the end in some scenarios, but I think that is the responsibility of the user to convert blanks to `NA` if they don't want them.  (Maybe that's too much effort to push to the user.)

Also, the use of `"_"` here relies on the fact that `sep = "_"`, but if I set `sep` to something else, it would not have the same type of behavior.

R/row_to_names.R
```diff
@@ -139,3 +146,9 @@ find_header <- function(dat, ...) {
   }
   ret
 }
+
+reduce_na <- function(x) {
```
billdenney, line 150: If the above suggestion to use `sapply` is correct, then this function is no longer needed.

billdenney (changes requested):
Since you said it's not yet ready to merge, I converted it to a draft PR.  I think you're on the right track, though tests make it easier to confirm.  I'm about to make some code review comments.

R/row_to_names.R
matanhakim: I agree, it was changed.

R/row_to_names.R
matanhakim: I agree, I changed the code to your suggestion.
To handle NA variable names i Used `stringr::str_replace_na()`, this is a classic use case for this.

R/row_to_names.R
matanhakim: It was removed.

Finally, I wanted to note that when using `remove_rows_above = TRUE`, I opted to start from the first row of the input in `row_number`.
Another idea would be to default to the topmost row, for example when the input is something like `row_number = 2:1`, but this is your decision to make, I guess.

R/row_to_names.R
```diff
@@ -59,7 +70,7 @@ row_to_names <- function(dat, row_number, ..., remove_row = TRUE, remove_rows_ab
       c()
     },
     if (remove_rows_above) {
-      seq_len(row_number - 1)
+      seq_len(row_number[1] - 1)
```
billdenney, line 73: I think that I'd go with removing everything above the lowest header row.  I.e. `seq_len(max(row_number) - 1)`.

tests/testthat/test-row-to-names.R
```diff
@@ -223,3 +248,34 @@ test_that("find_header works within row_to_names", {
     setNames(find_correct[4:nrow(find_correct),], c("D", "E"))
   )
 })
+
+test_that("multiple rows input works", {
+  q_row_to_names <- purrr::quietly(row_to_names)
```
billdenney, line 253: Rather than using `q_row_to_names`, can you please wrap the individual tests in `suppressMessages()`?  For my way of reading code, that makes tests simpler to understand.

billdenney (changes requested):
This looks pretty good to me.  @sfirke, what do you think?

tests/testthat/test-row-to-names.R
```diff
@@ -223,3 +248,34 @@ test_that("find_header works within row_to_names", {
...
+    c("Title", "Title2")
+  )
+  
+  expect_equal(
+    row_to_names(example_data_row_to_names[[1]], row_number=1:5) %>% 
+      names(),
+    c("Title_1_2_3", "Title2_4_5_6")
+  )
+  
+})
```
billdenney, line 281: Can you please add one more test for multiple NA values being combined and becoming a single "NA"?

billdenney (changes requested):

R/row_to_names.R
matanhakim: Done

tests/testthat/test-row-to-names.R
matanhakim: Done.
I went with `suppressWarnings()` since this was a warning rather than a message.

tests/testthat/test-row-to-names.R
matanhakim: I added another test, but there was also an existing test where I tried to do this without creating another data frame:

```
expect_equal(
    suppressWarnings(
      row_to_names(example_data_row_to_names[[1]], row_number=c(1,1)) %>% 
        names()
    ),
    c("NA", "NA")
  )
```

You can either keep both tests, or remove one of them to your liking.

billdenney: Thanks, I appreciate the new more explicit test.

billdenney:
Merged.  Thanks!

sfirke:
This looks great!  Sorry to miss the @ mention above.  Thanks so much @matanhakim for contributing, @billdenney for review, and both for getting it to an excellent place 🙏

matanhakim:
Thank you both for bringing this rich package to the community and facilitating my first PR feature contribution to FOSS in general and to the R community specifically🙏

billdenney:
And, @matanhakim, thanks for sticking with it to get the code into `janitor`!!!

PR #544, opened 2023-05-26
`get_one_to_one()` no longer errors with near-equal values that become identical factor levels

### LINKED ISSUE #543: `get_one_to_one()` errors with duplicated dttm
```r
library(fs)
janitor::get_one_to_one(fs::dir_info())
```
Please briefly describe your problem and what output you expect.

``` r
fs::dir_info() |> janitor::get_one_to_one()
#> Error in `levels<-`(`*tmp*`, value = as.character(levels)):   factor level [5] is duplicated
```

<sup>Created on 2023-05-26 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

This only seems to occur with duplicated dates, in this case with columns at position 5, 16, 17.

billdenney:
fix #543

This has the benefit that it simplifies some of the code for `get_one_to_one`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,8 @@
 
 * `make_clean_names()` no longer accepts a data.frame or tibble as input, use `clean_names()` for that (fix #532, **@billdenney**).
 
+* `get_one_to_one()` no longer errors with near-equal values that become identical factor levels (fix #543, thanks to @olivroy for reporting)
+
 # janitor 2.2.0 (2023-02-02)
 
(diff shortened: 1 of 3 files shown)


billdenney:
@sfirke, can you please take a look?

sfirke:
Yep will do

sfirke:
Indeed!  Gotta love a PR that simplifies the code base _and_ fixes a bug.  Thanks @billdenney  !

PR #545, opened 2023-05-27
Bug fix: single-column data frame input for `row_to_names()` now works

### LINKED ISSUE #536: Feature suggestion: allow multiple rows input to `row_to_names()`

matanhakim:
## Description
Before this PR and after merging #542, code like this would error:
```
library(janitor)
df <- data.frame(x = c("", "", "", "1"))

df |>
  row_to_names(1:3) |> 
  names()
```

this is because in the code:
```
new_names <- sapply(dat[row_number,], paste_skip_na, collapse = sep) %>% 
    stringr::str_replace_na()
```
the specific call `dat[row_number,]` turns the single-column data frame (i.e., a list of length 1) to a character vector (of length 3 in this case). This, in turn, makes `sapply()` to iterate over a wrong vector and therefore error.

Changing this to `dat[row_number, , drop = FALSE]` solves this.

A relevant test was added to prevent this in the future.

Finally, documentation for the changed behavior of the `row_number` argument in `row_to_names()` was added, and the NEWS file was edited to reflect changes in this PR and in #542.

## Related Issue
Fixes #536 and follows up on #542.

## Example
See above.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,8 +1,16 @@
 # janitor 2.2.0.9000 - unreleased development version
 
+## Breaking changes
+
+These are all minor breaking changes resulting from enhancements and are not expected to affect the vast majority of users.
+
+* When using `row_to_names()`, when all input values in `row_number` for a column are `NA`, `row_to_names()` creates a column name of `"NA"`, a character, rather than `NA`. If code previously used relied on a column name of `NA`, it will now error. To fix this, rely on a column name of `"NA"`.
+
 ## New features
 
 * A new function `paste_skip_na()` pastes without including NA values (#537).
+
+* `row_to_names()` now accepts multiple rows as input, and merges them using a new `sep` argument (#536). The default is `sep = "_"`. When handling multiple `NA` values, `row_to_names()` ignores them and only merges non-NA values for column names. When all values are `NA`, `row_to_names()` creates a column name of `"NA"`, a character, rather than `NA`.
 
 ## Bug fixes
 
--- a/R/row_to_names.R
+++ b/R/row_to_names.R
@@ -53,7 +53,7 @@ row_to_names <- function(dat, row_number, ..., remove_row = TRUE, remove_rows_ab
   if (is.na(sep)) {
     stop("`sep` can't be of type `NA_character_`.")
   }
-  new_names <- sapply(dat[row_number, ], paste_skip_na, collapse = sep) %>% 
+  new_names <- sapply(dat[row_number, , drop = FALSE], paste_skip_na, collapse = sep) %>% 
     stringr::str_replace_na()
   
   if (any(duplicated(new_names))) {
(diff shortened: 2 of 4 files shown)


billdenney:
This looks good.  Thanks!

PR #547, opened 2023-06-02
Remove most superseded dplyr functions

olivroy:
I removed the *_at, *_if, *_all functions from code and examples.

I replaced the example with another syntax.

For the tidygraph test, I did the change so that types are not coerced under the hood. Let me know if you think it is incorrect
`dplyr::mutate(dplyr::across(dplyr::where(is.numeric), ~ dplyr::coalesce(x, 1)))`

I can review.

Thanks for the great package!

A suggestion: maybe the package code could be restyled with styler to follow the tidyverse style guide! I'd be happy to contribute another PR if you think it's a good idea

PR #548, opened 2023-06-13
Restyle with `styler::style_pkg()`

olivroy:
Restyled the package with styler.

* Added a news bullet
* Replaced the `T`, `F` with `TRUE`, `FALSE`
* Whitespaces, indentation,  linebreaks
* Use `()` after pipes

See https://github.com/sfirke/janitor/pull/547#issuecomment-1584895857

In terms of automating the tidyverse style guide, there are 2 packages or actions that I found, but I am not very familiar with them so I would not feel comfortable implementing them in janitor.
* [lintr](lintr.r-lib.org)
* [precommit](https://lorenzwalthert.github.io/precommit/articles/precommit.html)


billdenney:
@sfirke, any thoughts?

tests/testthat/test-tabyl.R
```diff
@@ -296,7 +297,9 @@ test_that("NA levels get moved to the last column in the data.frame, are suppres
   expect_equal(names(y), c("down", "up", "NA_"))
   expect_equal(
     y[["NA_"]], # column c remains numeric
-    x %>% dplyr::filter(is.na(b)) %>% tabyl(c, a)
```
sfirke, line 299: It's interesting that it broke this pipeline up into multiple lines, there are others above that it left alone

tests/testthat/test-tabyl.R
```diff
@@ -391,7 +395,7 @@ test_that("3-way tabyl with 3rd var factor is listed in right order, #250", {
   z$cyl[32] <- NA
   expect_equal(names(tabyl(z, am, gear, cyl)), c("4", "8", "6", "NA_"))
   expect_equal(names(tabyl(z, am, gear, cyl, show_na = FALSE)), c("4", "8", "6"))
-  z <- z %>% dplyr::filter(! cyl %in% "4")
+  z <- z %>% dplyr::filter(!cyl %in% "4")
```
sfirke, line 398: Like here's one of many pipelines on one line that it left alone

sfirke (approved):
Just flagged two things that I wanted to understand how they are consistent.  But it looks good to me and IMO it's not worth modifying to try to address my curiosity :)

sfirke:
Agreed this is a nice boost to consistency.  I reviewed and in addition to what you noted Bill I saw a couple of equals-sign assignments replaced with `<-`.  Good stuff.  Let's merge it.

I do want to get something automated in place so that the repo can't get out of format and need a whopping big PR like this.  Thanks @olivroy for flagging precommit - that looks good to me.  The local install part seems a little odd, involving Python, but I gather we could set up [precommit lite](https://github.com/marketplace/actions/pre-commit-ci-lite) with a Github Action and it would make its own commits to PR branches that would run `styler` on the PR code.  I feel like that would be adequate for folks who don't want to mess with a local install of precommit?  The only part I don't understand from the docs is how to specify our action, i.e., run styler on all PRs.

sfirke:
That can be a separate thing though, no need to hold up this PR.

sfirke:
Hm some of these approaches to incorporating styler look reasonably easy: https://styler.r-lib.org/dev/articles/third-party-integrations.html  @olivroy are there any parameters you passed to styler or did you just use the default values?

sfirke:
I added styler as a GitHub action using the command `usethis::use_github_action("style")`.  And then pushed directly to main which sorry,I should have made it a PR 😳   https://github.com/sfirke/janitor/commit/41f08c89419a5e55078d7fc42106de46559aa98e

tests/testthat/test-tabyl.R
olivroy: according to the tidyverse style guide, 1 pipe is okay on a single line https://style.tidyverse.org/pipes.html#short-pipes

olivroy:
Hi, I used the default parameters. 

I wouldn't mind readjusting in some places if you'd prefer

tests/testthat/test-tabyl.R
sfirke: ah that was it! Thank you

PR #550, opened 2023-06-14
Remove the docs/ directory 

olivroy:
pkgdown site (already deployed via GitHub pages) last updated 3 years ago.

It is not necessary. I just wanted to test if it would work. tidyverse packages usually also don't have the docs/ directory tracked on GitHub.


sfirke:
Great catch that this directory is obsolete!

Issue #549, opened 2023-06-13
Upkeep proposition / spring cleaning


olivroy:
Hi,

I would have a couple suggestions to improve janitor

Let me know what you think and I can open PRs for the things you think are relevant!

* [x] Improve documentation
  *  Use `Roxygen: list(markdown = TRUE)` in `DESCRIPTION` to simplify the documentation (#551)
  *  (i.e. remove unnecessary explicit tags `@title`, `@description` etc. , `\{code}` and (#551)
  *  improve cross-references with `[fun]`  (#559) 
  * Use `@inheritParams` to avoid doc duplication (TODO later)
  * See https://r-pkgs.org/man.html
* [ ] Improve error, warnings and messages (use ~~`rlang::abort()`~~ or `cli::cli_abort()`and ~~use more snapshot tests to capture them~~ https://style.tidyverse.org/error-messages.html (#560)
* [x] Remove tidyr superseded functions (gather spread) (#552)

Edit:

* [x] Remove docs/ (#550)
* [x] Add favicon to the package website https://pkgdown.r-lib.org/reference/build_favicons.html (rename logo_small.png to logo.png) <https://r-pkgs.org/website.html#logo>. Retrieved .svg [here](https://github.com/maxogden/hexbin/blob/7a67b003156ccb8b20d46acfe72e1c507104eb25/vector/janitor.svg) #553 
* [x] Remove references to master branch (in README for example) (#551)
* [ ] Edit `_pkgdown.yml` so that functions are organized on the site https://sfirke.github.io/janitor/reference/index.html (see https://dplyr.tidyverse.org/reference/index.html for an example) 
* [x] Add `@keywords internal` to make sure hard deprecated functions don't show up in the package reference (#551)
* [x] Remove `index.md` and make README.md to website homepage (#554)

billdenney:
Thanks for this help!

I like all of this except for the snapshot tests.  I don't prefer snapshot tests because they don't fail on CRAN.  I prefer explicit tests for messages with regular expressions or specific examples so that we always know that we are testing exactly what we intend to test.

For the error messages, I'd prefer using `rlang` rather than `cli` since we already depend on `rlang` (even though I like the `cli` visuals better).

I definitely think we should remove the superseded function use, if reasonable.  We will need to be careful of functional changes when switching to the new functions.  `janitor` is downloaded >100k times/month, so breaking changes should be minimized at (almost) all costs.

@sfirke, what are your thoughts?

sfirke:
[...]

PR #552, opened 2023-06-19
Remove `tidyr::spread()`

olivroy:
## Description

Removing mentions of the superseded tidyr functions.

## Related Issue

Addresses #549

### DIFF
--- a/R/tabyl.R
+++ b/R/tabyl.R
@@ -209,7 +209,7 @@ tabyl_2way <- function(dat, var1, var2, show_na = TRUE, show_missing_levels = TR
   result <- tabl %>%
     tidyr::pivot_wider(
       names_from = !!var2,
-      values_from =  "tabyl_2way_n",
+      values_from = "tabyl_2way_n",
       values_fn = ~ dplyr::coalesce(.x, 0L),
       names_sort = TRUE
     )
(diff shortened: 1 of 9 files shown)


R/tabyl.R
```diff
@@ -207,7 +207,12 @@ tabyl_2way <- function(dat, var1, var2, show_na = TRUE, show_missing_levels = TR
   tabl[2][is.na(tabl[2])] <- "NA_"
   tabl[2][tabl[2] == ""] <- "emptystring_"
   result <- tabl %>%
-    tidyr::spread(!!var2, "tabyl_2way_n", fill = 0)
+    tidyr::pivot_wider(
+      names_from = !!var2,
+      values_from =  "tabyl_2way_n",
+      values_fn = ~ dplyr::coalesce(.x, 0L),
+      names_sort = TRUE
+    )
```
olivroy, line 215: From `tidyr::spread()` documentation, fill = 0 had to be replaced with  dplyr::coalesce(.x, 0L) to make sure implict and explicit missing values were replaced.
Also, `spread()` sorted names automatically, not pivot_wider(), hence the names_sort argument.

```
@param fill If set, missing values will be replaced with this value. Note that there are two types of missingness in the input: explicit missing values (i.e. NA), and implicit missings, rows that simply aren't present. Both types of missing value will be replaced by fill.
```

sfirke:
Excellent!  Thank you for commenting the most complex section.  And nice job catching the instance of "spread" that is in a comment, not code 😁

R/tabyl.R
sfirke: thanks for explaining this part!

sfirke:
BTW did the auto-linter run catch the whitespaces that prompted your last commit?

olivroy:
Hi, sfirke, it did, in the sense that I get a notification that the style action didn't pass. See https://github.com/sfirke/janitor/issues/549#issuecomment-1592931661

Issue #386, opened 2020-07-03
Ordered factors and 3-way tables in Janitor


HenkvdKolk:
With a dataframe with three *ordered factors*, Janitor does not produce a 2x2x2 table. With unordered factors there seems to be no problem. Why is that? The example is with numbers because of the rbinom I used, but the output is the same. Is seems that the ordered nature of x is the biggest problem, with a simple factor I get what I need. This is problematic with low/middle/high type of factors.

```r
library(tidyverse)
library(janitor)

n <- 100
data <- rbinom(n, 1, 0.5)  %>% factor(order = TRUE, levels = c("0", "1")) %>% as_tibble() %>% rename(., x = value)
data$y <- rbinom(n, 1, 0.5) %>% factor(order = TRUE, levels = c("0", "1"))
data$t <- rbinom(n, 1, 0.5) %>% factor(order = TRUE, levels = c("0", "1"))

data %>% 
  tabyl(x, y, t, show_na = FALSE) %>% 
  adorn_totals("col") %>% 
  adorn_totals("row") %>% 
  adorn_percentages("col") %>% 
  adorn_pct_formatting(digits = 1) %>%
  adorn_ns() #show the numbers of cases
```

sfirke:
Thank you for this bug report!  I apologize for not realizing what you were saying, that this command errors (I didn't run the code).  I will aim to get this fixed in the next week or so.

HenkvdKolk:
Thanks. I will frame this e-mail as this was my first successful bug-report. I like the package (the tabyl part at least), so also thanks for creating it.

sfirke:
Fixed :grin:  Thanks again for this bug report!

PR #561, opened 2023-09-10
Minimal argument checking review

olivroy:
## Description
Standardize argument checking with `rlang::arg_match0()` + use inherits instead of class(x) %in%

## Related Issue
Addresses #549 

## Example

See updates in tests, but this just uses https://rlang.r-lib.org/reference/arg_match.html + inherits in some places.

### DIFF
--- a/R/tabyl.R
+++ b/R/tabyl.R
@@ -247,7 +247,7 @@ tabyl_3way <- function(dat, var1, var2, var3, show_na = TRUE, show_missing_level
   # grab class of 1st variable to restore it later
   col1_class <- class(dat[[1]])
   col1_levels <- NULL
-  if ("factor" %in% col1_class) {
+  if (is.factor(dat[[1]])) {
     col1_levels <- levels(dat[[1]])
   }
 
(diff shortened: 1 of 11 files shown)


R/tabyl.R
```diff
@@ -247,7 +247,7 @@ tabyl_3way <- function(dat, var1, var2, var3, show_na = TRUE, show_missing_level
   # grab class of 1st variable to restore it later
   col1_class <- class(dat[[1]])
   col1_levels <- NULL
-  if ("factor" %in% col1_class) {
+  if (is.factor(dat[[1]])) {
```
olivroy, line 250: For this change, I tried to see if there was a specific reason why is.factor was not used. but I couldn't find any notes.

```r
is.factor(ordered(c("1", "2")) 
#> TRUE
```

This line seems to have been implemented in #176, but no note on why is.factor was not used. and was fixed in #386 to what it is today.

olivroy:
@billdenney in case you have time to look at this one! it is fairly minimal and straightforward!

billdenney:
This looks good to me.

PR #569, opened 2024-03-02
Clarify documentation of update remove_empties.R

### LINKED ISSUE #568: `cutoff` argument in `remove_empty()` is being implemented confusingly
In `remove_empty()`, the documentation of the `cutoff` argument states:

> What fraction (>0 to <=1) of rows or columns must be **empty** to be removed?

The bold on "empty" is mine. This suggests that the **higher** `cutoff` is, **more** rows are supposed to be kept, since the condition to remove them is stricter.

Nevertheless, when you use different values of `cutoff`, the opposite behavior is observed; the **higher** `cutoff` is, **less** rows are kept. Please see the reprex for evidence.

I see two optional solutions:

1. Change the behavior of `cutoff`, so that it would behave as described in the documentation. This has the major drawback of being a drastically breaking change. The source for this behavior lies in rows [60](https://github.com/sfirke/janitor/blob/3769ab0ee67d60d9f0533fcdf0b9bc2b6b63f4f5/R/remove_empties.R#L60) and [74](https://github.com/sfirke/janitor/blob/3769ab0ee67d60d9f0533fcdf0b9bc2b6b63f4f5/R/remove_empties.R#L74) of `remove_empty.R`, and a possible solution might be to drop `!` from the calculation.
2. Change the documentation of `cutoff`, so it would be clear what behavior is expected.

Happy to hear your thoughts on this!

``` r
library(tidyr)
library(janitor)
#> 
#> Attaching package: 'janitor'
#> The following objects are masked from 'package:stats':
#> 
#>     chisq.test, fisher.test

billboard |>
  nrow()
#> [1] 317

billboard |> 
  janitor::remove_empty("rows", cutoff = 0.9) |> 
  nrow()
#> [1] 0

billboard |> 
  janitor::remove_empty("rows", cutoff = 0.1) |> 
  nrow()
#> [1] 293
```

<sup>Created on 2024-02-26 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

matanhakim:
Fixes #568 

Feel free to edit or change, it is a bit difficult to write a clear description.
Added a clarification for lower and higher values.

### DIFF
--- a/R/remove_empties.R
+++ b/R/remove_empties.R
@@ -7,7 +7,7 @@
 #' @param which one of "rows", "cols", or `c("rows", "cols")`. Where no
 #'   value of which is provided, defaults to removing both empty rows and empty
 #'   columns, declaring the behavior with a printed message.
-#' @param cutoff Under what fraction (>0 to <=1) of non-empty rows or columns should
+#' @param cutoff a row/col should have at least this fraction (>0 to <=1) of non-NA values to be
 #'   `which` be removed? Lower values keep more rows/columns, higher values drop more.
 #' @param quiet Should messages be suppressed (`TRUE`) or printed
 #'   (`FALSE`) indicating the summary of empty columns or rows removed?
--- a/R/remove_empties.R
+++ b/R/remove_empties.R
@@ -8,7 +8,7 @@
 #'   value of which is provided, defaults to removing both empty rows and empty
 #'   columns, declaring the behavior with a printed message.
 #' @param cutoff a row/col should have at least this fraction (>0 to <=1) of non-NA values to be
-#'   `which` be removed? Lower values keep more rows/columns, higher values drop more.
+#'   retained.  E.g., `cutoff = 0.8` means that rows/cols that are 20% or more missing will be dropped.
 #' @param quiet Should messages be suppressed (`TRUE`) or printed
 #'   (`FALSE`) indicating the summary of empty columns or rows removed?
 #' @return Returns the object without its missing rows or columns.
--- a/R/remove_empties.R
+++ b/R/remove_empties.R
@@ -7,8 +7,8 @@
 #' @param which one of "rows", "cols", or `c("rows", "cols")`. Where no
 #'   value of which is provided, defaults to removing both empty rows and empty
 #'   columns, declaring the behavior with a printed message.
-#' @param cutoff What fraction (>0 to <=1) of rows or columns must be empty to
-#'   be removed?
+#' @param cutoff Under what fraction (>0 to <=1) of non-empty rows or columns should
+#'   `which` be removed? Lower values keep more rows/columns, higher values drop more.
 #' @param quiet Should messages be suppressed (`TRUE`) or printed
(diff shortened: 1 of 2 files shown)


billdenney:
The update looks good.  Can you please update the full documentation (`devtools::document()`) so that everything is consistent and push that update?

billdenney:
This looks good to me!

R/remove_empties.R
```diff
@@ -7,8 +7,8 @@
 #' @param which one of "rows", "cols", or `c("rows", "cols")`. Where no
 #'   value of which is provided, defaults to removing both empty rows and empty
 #'   columns, declaring the behavior with a printed message.
-#' @param cutoff What fraction (>0 to <=1) of rows or columns must be empty to
-#'   be removed?
+#' @param cutoff Under what fraction (>0 to <=1) of non-empty rows or columns should
```
sfirke, line 10:
```suggestion
#' @param cutoff a row/col should have at least this fraction (>0 to <=1) of non-NA values to be
```

R/remove_empties.R
```diff
@@ -7,8 +7,8 @@
...
+#'   `which` be removed? Lower values keep more rows/columns, higher values drop more.
```
sfirke, line 11:
```suggestion
#'   retained.  E.g., `cutoff = 0.8` means that rows/cols that are 20% or more missing will be dropped.
```

R/remove_empties.R
```diff
@@ -7,8 +7,8 @@
 #' @param which one of "rows", "cols", or `c("rows", "cols")`. Where no
 #'   value of which is provided, defaults to removing both empty rows and empty
 #'   columns, declaring the behavior with a printed message.
-#' @param cutoff What fraction (>0 to <=1) of rows or columns must be empty to
-#'   be removed?
+#' @param cutoff a row/col should have at least this fraction (>0 to <=1) of non-NA values to be
```
sfirke, line 10:
```suggestion
#' @param cutoff a row/col should have at above this fraction of non-NA values to be
```

R/remove_empties.R
```diff
@@ -7,8 +7,8 @@
 #' @param which one of "rows", "cols", or `c("rows", "cols")`. Where no
 #'   value of which is provided, defaults to removing both empty rows and empty
 #'   columns, declaring the behavior with a printed message.
-#' @param cutoff What fraction (>0 to <=1) of rows or columns must be empty to
-#'   be removed?
+#' @param cutoff a row/col should have at above this fraction of non-NA values to be
```
sfirke, line 10:
```suggestion
#' @param cutoff a row/col should have more than this fraction of non-NA values to be
```

sfirke:
This is a tricky one to describe in a brief and clear way :sweat_smile: @matanhakim and @billdenney what do you think of my attempt that built on Matan's - anything still to improve?

billdenney (approved):
It looks good to me with the current edits.

matanhakim:
This looks great to me. Indeed a tricky explanation!

Issue #539, opened 2023-04-27
Feature suggestion: `most()` and `assert_count_true()`


billdenney:
I sometimes get dirty data that has multiple values that I need to choose one from.  In a recent example, I received a dataset where an individual had multiple values for their sex (both male and female when they definitely did not undergo gender reassignment between the measurements).

To work with these types of issues, I think that two different types of functions can help:

`most(x)` is a companion to `any()` and `all()` from base R.  It takes in a vector, `x`, and returns true if more than half of the values are `isTRUE(x)`.

`assert_count_true(x, n)` takes in a logical vector `x` and an expected count that should be `isTRUE(x)`, `n`.  If `sum(isTRUE(x)) == n`, then it returns `x`.  If a different number are `TRUE`, then it returns an error indicating the mismatch in count.

billdenney:
These are different from some of the typical `janitor` functions, so I'd like @sfirke to weigh in on if they feel like a good fit.

sfirke:
I'm fine with adding `most()`.  In your example, might you use it like:
```
dat %>%
  group_by(id) %>%
  filter(most(gender == "male"))
```
To get the data for all participants for whom most of their `gender` values are `male` ?  Just checking my understanding.

Should it take an option cutoff value that defaults to `0.5`?  Then maybe we would be talking about calling it `at_least()` ...

Not trying to muddy the waters, just want to get precise on design and use cases.

sfirke:
Could you share an example of using `assert_count_true()`?  With `mtcars` or similar?  I don't quite grasp how I would use it.  There have been some talks about assertive checks in janitor, I can't remember what approach we landed on, but in general I support them.  `janitor::compare_df_cols_same` is made for assertion.

billdenney:
I like `at_least(x, fraction = 0.5)` more than `most()` as it covers a more general case with no more user effort.

For `assert_count_true()`, I often use something like it in my data cleaning routines.  I have data where I know that one particular row is bad.  I want to make sure that I only match that one row and no more or fewer (or maybe 5 rows or...).  My use case looks like:

```r
cleaned_data <-
  data |>
  mutate(
    col_to_update =
      case_when(
        assert_count_true(Person == "Bill" & Age == 40, count = 1) ~ 29, # The fountain of youth :)
        TRUE ~ col_to_update
      )
  )
```

My implementation looks something like (if I'm using `deparse()` correctly-- the idea is to tell the user the actual called value for `x`):

```r
[...]

PR #573, opened 2024-05-22
Add `assert_count_true()` to verify that an expected number of values are `TRUE`

billdenney:
Related to #539

I use a variant of this often in my data pipelines where I want to select a certain number of values and be sure that I'm only getting the number I expect.  For example, I want to change one date for one person, and I want to be sure that I only change that one date:

```
# We are very young
my_data <- data.frame(name = c("Bill", "Sam"), birthdate = c("2024-05-22", "2024-05-22"))
my_data |>
  dplyr::mutate(
    birthdate =
      dplyr::case_when(
        assert_count_true(name == "Bill" & birthdate == "2024-05-22") ~ "2024-05-23",
        TRUE ~ birthdate
      )
  )
```


sfirke:
@billdenney thanks for this and sorry to take so long to review!

Three things come to mind after reviewing:

**1. Function name**
Would it be cleaner to call this just `assert_count` ? I think the `true` is clear enough from the usage, but I don't feel strongly if you prefer the current way.

**2. Providing a before-and-after example**

I get the value of this function and would use it. But it took me reading each code block to understand its value. I think we can make it clearer upfront in the documentation (package vignette and function docs) that this doesn't _change_ anything in the data; it's a wrapper that performs a data quality check mid-pipeline and halts things if an assertion fails.

The best way I can think of is to show a scenario with before and after example.  For instance:

------
Imagine we have a column in a dataset that contains student test scores.  A few scores were entered into the grade platform after the data file was exported, so we're filling them in manually during data cleaning. Ordinarily we might do that like this:
```
clean <- 
  raw |>
  dplyr::mutate(
    test_score =
      case_when(
        student_id == 12345 & is.na(test_score) ~ 84,
        student_id == 67890 & is.na(test_score) ~ 91,
        TRUE ~ test_score
      )
  )
```
But this has a risk: if in a later version of the data, one of these students is missing two test scores, the value will be imputed for both cases. `assert_count_true` defends against that. It wraps the statements, doing nothing if the assumed count is matched and throwing an error if not.

```
clean <- 
  raw |>
  dplyr::mutate(
    test_score =
      case_when(
        assert_count_true(student_id == 12345 & is.na(test_score), count = 1) ~ 84,
        assert_count_true(student_id == 67890 & is.na(test_score), count = 1) ~ 91,
        TRUE ~ test_score
      )
  )
```
------
**3. janitor's approach in wading into the assertion space**
I may be overthinking a little here but this feels like the start of a separate chapter, even if it's only one case for now. One implication is that I think we could carve out a little separate space in terms of function organization and documentation. Like this could go in a file `assertions.R`.

I am also considering how janitor complements the existing assertion functions out there. I looked at the [assertr vignette](https://cran.r-project.org/web/packages/assertr/vignettes/assertr.html) and [this blog post on 4 ways to be assertive in R](https://blog.djnavarro.net/posts/2023-08-08_being-assertive/).

To me the big difference is that `assert_count_true` does not take its own line in the pipeline, it wraps an existing line.  I didn't see that in the other packages. You could make these same checks with `assertr::assert()` but at a cost of needing to write two lines for each replacement case: one to assert-check, one to replace. With your function, it's one line for each replacement, so there's less code to read and maintain.

I would argue this niche is unfilled as far as I know and that's this function's reason for being. Do you agree? If so maybe we briefly note that in the function or package docs so people understand when/why to use it.

billdenney:
@sfirke:

For point 2, yes, I like the example. I will incorporate it in some way (maybe a new assertion vignette).

For points 1 and 3, you understand all the ideas that I'm looking toward. I was thinking of the `checkmate` library which uses `assert_*` to ensure that aspects of input arguments are true, and Danielle Navarro's blog post (aside: she has an amazing blog!) shows other ways that this makes sense. I thought that it could be a good fit for `janitor` specifically because of my typical use case which you succinctly described in point 2. I use it to clean dirty data that may change over time or that I want to be sure I only capture the expected number of rows for, even if it doesn't change over time.

I agree that this would be a new chapter for `janitor`, and it deserves a new organizational layer. I'll work those in over the next few days. I also agree that the `_true` part of the function name can be assumed (and therefore removed from the function name).

billdenney:
@sfirke, I think that I covered all the to-do items here. Please take a look.

Also, I found a bug in either the `commonmark` or `xml2` package while trying to add spell checking to the package. So, spell checking is only partially added (that's unrelated to the main purpose of the PR, but it seemed helpful sorry for making this a bigger change to review).

billdenney:
I merged the current `main` branch back in here. This should be ready to review again. When it's merged, we can head over to the bigger package separation PR.

billdenney:
@sfirke, ping on this one. I think it's ready.

Issue #563, opened 2024-01-21
Set old names as labels in `clean_names`


jospueyo:
Today I engaged in this thread in [mastodon](https://fosstodon.org/@bbolker@fediscience.org/111791764736816890) who led me to read this [blog](https://www.pipinghotdata.com/posts/2022-09-13-the-case-for-variable-labels-in-r/) about data labels in R. I thought that would be nice that `clean_names()` had the option to keep the old names as labels in the dataframe.

I think that would not be too complicated. I'm keen to propose a PR if I you were willing to accept the feature. It would be something like this...  

```r

if (set_labels){
    old_names <- names(dat)
    purrr::walk(1:ncol(dat), \(i) attr(dat[[i]], "label") <<- old_names[[i]])
}

names(dat) <- make_clean_names(names(dat))

 
```

If you don't like using superassignment, this can also be done in a for loop.

billdenney:
@jospueyo , thanks for your interest in the package.  And, I understand the value of keeping the original name associated with the column.

I like that it seems you would make including the labels as optional with the `set_labels` argument.  I also like that you accurately predicted that I would not like super-assignment. 😄 

So, I think that PR could be useful.  As with any PR:

* Please make the default backwards compatible (so `set_labels` would be `FALSE`.  I, too, have had issues with labels messing things up in the past, as mentioned on the Mastodon thread.
* Please be sure to test for the changes
* please add it to the News file
* please add yourself as a contributor

sfirke:
Sounds good to me too. After this is introduced I could see having tabyl() use labels where available, but that's unrelated (in code at least) to this suggestion.

sfirke:
Looking at that Masto thread, some people talk about wanting to restore the original names at some point.  Do we know if there's a function in another labelling package that transfers the labels into the column names?  Kind of undoing the `clean_names` call.

It seems like a potentially simple function, but (a) I hope someone else already wrote a good version we can point to (b) I don't want to overcomplicate things and there would be some thinking needed around say, what if only some of the original columns are present.

billdenney:
Something like `dirty_names()`? 🤣 (I don't know if it already exists.)

My experience with attributes like this is that they're lost easily.  Creating a reverse function (like `dirty_names()` or likely better `label_to_names()`) would likely end with a lot of questions.  That doesn't mean we shouldn't create it, but we should have some very good documentation before including it in `janitor`.

sfirke:
[...]

Issue #394, opened 2020-07-25
add labels to varnames/values as in {labelled} and {haven}


higgi13425:
## Feature request

Recognize variable labels and value labels as in {haven} 
[](https://haven.tidyverse.org)

and {labelled} 
[](https://larmarange.github.io/labelled/index.html)

packages when printing tabyls.

It would be cool.

sfirke:
I like this idea and in fact I have a basic version that I use for tabyl-ing labelled vectors, back when that was a data type I encountered more: https://github.com/sfirke/tidysurvey/issues/2#issuecomment-463262151

How does that work for you?  It is just a minimal example I used in my own work and hasn't received the thought it would need to be package-ready.

sda030:
I strongly support recognizing {labelled} in tabyls. I am about to introduce tidyverse, janitor and more to many R-newbies who face a lot of survey data, but this limitation makes me reconsider tabyl... :(

olivroy:
@sfirke I could implement a PR for this.. Display labels in `tabyl()` if they exist.

New support for the label attribute was recently added in RStudio. gt does this, and ggplot2 seems to be going in this direction tidyverse/ggplot2#5879

sfirke:
@olivroy that would be great.  I will review such a PR.  Let me know if you have design questions as you go, or questions about the backend code of `tabyl`.

PR #564, opened 2024-01-21
add set_labels to clean_names()

jospueyo:
Implements #563.

I added the `set_labels` argument ensuring backward compatibility for default (not for dimnames) and sf methods. I'm not confident enough with tidygraph to implement the method for that.

I did not create labels when dimnames is used because I thought when dimnames is used, `dat` is probably a matrix, which doesn't accept column labels, am I right?

I tested for both methods.

I added the new feature in NEWS and myself as a contributor in the DESCRIPTION (thanks for this).

Also, I fixed a note regarding a bad header in NEWS (line 27).

Issue #527, opened 2023-02-03
Adorn_totals() working only with tibbles?


larry77:
Hello,
In my workflow I often call tabyl on some data and as a consequence I get an object having both "tabyl" and "data.frame" class.
It seems to me that if the resulting object inherits the "tabyl" class then adorn_totals will fail on it.
Why? I think I did not have this in the past.
In the reprex below, df has only the "data.frame" class and everything is fine.
df2 has also the tabyl class (in reality it comes from calling tabyl on some data) and then adorn_totals() fails.
Any idea if this is intended?

Thanks!

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(janitor)
#> 
#> Attaching package: 'janitor'
#> The following objects are masked from 'package:stats':
#> 
#>     chisq.test, fisher.test

df <- structure(list(procedure_name = c("Agriculture Block Exemption Regulation", 
"Fisheries Block Exemption Regulation", "General Block Exemption Regulation", 
"Notified Aid"), n = c(38L, 1L, 215L, 51L), percent = c(12.5, 
0.3, 70.5, 16.7)), row.names = c(NA, -4L), class = c(## "tabyl", 
                                                     "data.frame"))

df
#>                           procedure_name   n percent
#> 1 Agriculture Block Exemption Regulation  38    12.5
#> 2   Fisheries Block Exemption Regulation   1     0.3
#> 3     General Block Exemption Regulation 215    70.5
#> 4                           Notified Aid  51    16.7

is.data.frame(df)
#> [1] TRUE

df |>
    adorn_totals()
#>                          procedure_name   n percent
#>  Agriculture Block Exemption Regulation  38    12.5
#>    Fisheries Block Exemption Regulation   1     0.3
#>      General Block Exemption Regulation 215    70.5
#>                            Notified Aid  51    16.7
#>                                   Total 305   100.0

df2 <- structure(list(procedure_name = c("Agriculture Block Exemption Regulation", 
"Fisheries Block Exemption Regulation", "General Block Exemption Regulation", 
"Notified Aid"), n = c(38L, 1L, 215L, 51L), percent = c(12.5, 
0.3, 70.5, 16.7)), row.names = c(NA, -4L), class = c( "tabyl", 
                                                     "data.frame"))


df2 |>
[...]

PR #575, opened 2024-05-23
Tabyl label

olivroy:
## Description

Addresses https://github.com/sfirke/janitor/issues/563#issuecomment-1902660518

SImply default to label attribute if it exists when displaying tabyl.

It enables this by default, which is a breaking change? But I wouldn't see why you would not want that if you took care of labelling your variables.

More fiddling required than expected since a lot of base R functions drop attributes `subset()`, `split()`, `as.character()`, etc.

A solution to get the previous behavior is the following (or haven::zap_labels() or similar.

```r
tabyl(mt_label %>% labelled::remove_var_label(), cyl)
```

## Related Issue

Partially addresses #394. Doesn't solve the value labels part, only variable label.

## Example

```r
  mt_label <- mtcars
  attr(mt_label$cyl, "label") <- "Number of cyl"
# before 
tabyl(mt_label, cyl)
 cyl  n percent
   4 11 0.34375
   6  7 0.21875
   8 14 0.43750
# this pr 
tabyl(mt_label, cyl)
 Number of cyl  n percent
             4 11 0.34375
             6  7 0.21875
             8 14 0.43750
```


R/tabyl.R
```diff
@@ -250,6 +257,10 @@ tabyl_3way <- function(dat, var1, var2, var3, show_na = TRUE, show_missing_level
   dat <- dplyr::select(dat, !!var1, !!var2, !!var3)
   var3_numeric <- is.numeric(dat[[3]])
 
+  # Sometimes, attributes can be dropped with transformation.
```
sfirke, line 260:
```suggestion
  # Preserve labels, as attributes are sometimes dropped during transformations.
```

R/tabyl.R
```diff
@@ -277,7 +288,14 @@ tabyl_3way <- function(dat, var1, var2, var3, show_na = TRUE, show_missing_level
     dat[[2]] <- as.factor(dat[[2]])
   }
 
-  result <- split(dat, dat[[rlang::quo_name(var3)]]) %>%
+  result <- split(dat, dat[[rlang::quo_name(var3)]])
+  # split() drops attributes, so we manually add back the 1st variable attribute.
```
sfirke, line 292:
```suggestion
  # split() drops attributes, so we manually add back the label attributes.
```
I was confused by "1st" here, since this adds back two labels -- or is 1st correct and I'm missing something?

sfirke (approved):
Outstanding!  Easy to follow, thank you.  I just suggested teeny comment changes.

R/tabyl.R
olivroy: Thanks for the review. I meant initial label attribute. Your review comment captures it correctly!

sfirke:
This is great @olivroy, sorry for delay in review.  When we merge this we can see if the Codecov error #576 is fixed 😁 

> It enables this by default, which is a breaking change? But I wouldn't see why you would not want that if you took care of labelling your variables.

Agree and agree... the next release of janitor should probably be 3.0.0 given that this is a breaking change and that so much time has elapsed.  I do expect it could cause pain even if it's an improvement if it breaks existing code.

> More fiddling required than expected since a lot of base R functions drop attributes subset(), split(), as.character(), etc.

Yes this is irritating.  #527 would at least address the tidyverse functions doing this, I think.

> A solution to get the previous behavior is the following (or haven::zap_labels() or similar.

Could you add that to `NEWS.md` in this PR and move this contribution from "New features" to "Breaking changes"?

NEWS.md
```diff
@@ -8,6 +8,8 @@ These are all minor breaking changes resulting from enhancements and are not exp
 
 ## New features
 
+* `tabyl()` now defaults to displaying the label attribute for the column name (@olivroy, #394).
+
```
olivroy, line 12:
```suggestion

* `tabyl()` now defaults to displaying the label attribute for the column name (@olivroy, #394).

## New features

```

olivroy:
Thanks so much for the review and the kind words! I am very glad my code was easy to follow, but I have to thank you and Bill for that. janitor is one of the first packages I contributed to!

It is a breaking change, but only in display. I don't think many people will pipe much after. And IIRC, the original names are kept in tabyl.

NEWS.md
```diff
@@ -6,6 +6,8 @@ These are all minor breaking changes resulting from enhancements and are not exp
 
 * When using `row_to_names()`, when all input values in `row_number` for a column are `NA`, `row_to_names()` creates a column name of `"NA"`, a character, rather than `NA`. If code previously used relied on a column name of `NA`, it will now error. To fix this, rely on a column name of `"NA"`.
 
+* `tabyl()` now defaults to displaying the label attribute for the column name (@olivroy, #394).
```
olivroy, line 9:
```suggestion
* `tabyl()` now defaults to displaying the label attribute for the column name (@olivroy, #394). To revert to previous behavior (ignore variable labels, you can use `haven::zap_labels()`, `labelled::remove_labels()` or similar, before piping in `tabyl()`.

```

olivroy:
This will work well with #564 as in my experience, names when you read from Excel in particular tend to be more human readable, the ability to keep those and then use them automatically for `tabyl()` is an improvement in my opinion!

NEWS.md
```diff
@@ -6,6 +6,9 @@ These are all minor breaking changes resulting from enhancements and are not exp
 
 * When using `row_to_names()`, when all input values in `row_number` for a column are `NA`, `row_to_names()` creates a column name of `"NA"`, a character, rather than `NA`. If code previously used relied on a column name of `NA`, it will now error. To fix this, rely on a column name of `"NA"`.
 
+* `tabyl()` now defaults to displaying the label attribute for the column name (@olivroy, #394). To revert to previous behavior (ignore variable labels, you can use `haven::zap_labels()`, `labelled::remove_labels()` or similar, before piping in `tabyl()`.
```
sfirke, line 9:
```suggestion
* When `tabyl()` is called on a data.frame containing labels, it now displays the label attribute as the name of the first column in the the resulting `tabyl` object (@olivroy, #394). This may break subsequent code that refers to the output of such a `tabyl` by column name. To maintain the previous behavior of ignoring variable labels, you can remove the labels with a function like `haven::zap_labels()` or `labelled::remove_labels()` before calling `tabyl()`.
```

PR #579, opened 2024-09-12
Fix issue where the sf package "sf_column" may not be the last column in an sf object

billdenney:
@ar-puuk, can you please test this and confirm that it works for you.  I'm not an `sf` package user, so I can just confirm that it prints correctly not that the rest of it works.


billdenney:
@sfirke, I think that this is ready to merge unless you have any comments on review.

sfirke (approved):
LGTM!

sfirke:
Thanks Bill and Pukar!

PR #584, opened 2024-11-18
Remove timezones no longer supported by Debian

### LINKED ISSUE #582: Submit to CRAN to fix Debian check failure
From CRAN today 11/18:

```
Dear maintainer,

Please see the problems shown on
<https://cran.r-project.org/web/checks/check_results_janitor.html>.

The errors on the Debian check systems are from a recent system upgrade
to tzdata 2024b which did

   Names present only for compatibility with UNIX System V
   (last released in the 1990s) have been moved to 'backward'.

which includes CET, CST6CDT, EET, EST*, HST, MET, MST*, PST8PDT, and
WET.

Debian ships the names in 'backward' in a separate package tzdata-legacy
which is not installed "by default".

The comments in the tzdata 'backward' file say

# Although this file is optional and tzdb will work if you omit it by
# building with 'make BACKWARD=', in practice downstream users
# typically use this file for backward compatibility.

so clearly one cannot unconditionally assume that the backward
compatibility names will work.

Can you please change your code (typically tests) so that they do not
fail when the backward compatibility names are not available?

Please correct before 2024-12-09 to safely retain your package on CRAN.

Best wishes,
The CRAN Team
```

### LINKED ISSUE #583: Remove timezone support from `sas_numeric_to_date()`
While working on #582, there was an issue with timezone tests in `sas_numeric_to_date()`.

Looking at https://communities.sas.com/t5/Ask-the-Expert/Tips-and-Tricks-for-Working-with-Dates-and-Times-in-SAS-Q-amp-A/ta-p/813974, it appears that SAS likely (but not definitively) doesn't support storing time zones and it likely stores the timezone as the current time zone. The documentation that I can find is not clear.

Therefore, I plan to remove timezone support from the function and always use UTC timezones. That seems like the least-bad choice because a SAS file sent to a computer in a timezone that has different daylight saving time method would likely get different results, based on my reading.

billdenney:
fix #582
fix #583 (rather than removing it, I warned about using timezones-- let the user make a choice)

### DIFF
--- a/R/sas_dates.R
+++ b/R/sas_dates.R
@@ -46,7 +46,7 @@ sas_numeric_to_date <- function(date_num, datetime_num, time_num, tz = "UTC") {
     has_datetime <- TRUE
   }
   if (has_datetime) {
-    ret <- as.POSIXct(datetime_num, origin = "1960-01-01", tz = "UTC")
+    ret <- as.POSIXct(datetime_num, origin = "1960-01-01", tz = tz)
   } else if (has_date) {
     ret <- as.Date(date_num, origin = "1960-01-01")
   } else if (has_time) {
--- a/R/sas_dates.R
+++ b/R/sas_dates.R
@@ -16,11 +16,16 @@
 #' sas_numeric_to_date(time_num = 3600) # 01:00:00
 #' @family date-time cleaning
 #' @export
-sas_numeric_to_date <- function(date_num, datetime_num, time_num, tz = "") {
+sas_numeric_to_date <- function(date_num, datetime_num, time_num, tz = "UTC") {
   # Confirm that a usable set of input arguments is given
   has_date <- !missing(date_num)
   has_datetime <- !missing(datetime_num)
   has_time <- !missing(time_num)
+  stopifnot(is.character(tz))
+  stopifnot(length(tz) == 1)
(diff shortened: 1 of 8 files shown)


R/sas_dates.R
```diff
@@ -37,9 +42,11 @@ sas_numeric_to_date <- function(date_num, datetime_num, time_num, tz = "") {
...
       stop("The same values are not NA for both `date_num` and `time_num`")
     }
-    ret <- as.POSIXct(86400 * date_num + time_num, origin = "1960-01-01", tz = tz)
-  } else if (has_datetime) {
-    ret <- as.POSIXct(datetime_num, origin = "1960-01-01", tz = tz)
+    datetime_num <- 86400 * date_num + time_num
+    has_datetime <- TRUE
+  }
+  if (has_datetime) {
+    ret <- as.POSIXct(datetime_num, origin = "1960-01-01", tz = "UTC")
```
sfirke, line 49: Is the `tz = "UTC"` here intentional instead of `tz = tz` ? Then the input value of `tz` won't be used? It sounds like in #583 that was initially your approach, then you clarified that you want to keep existing code working.

billdenney: Thanks for the review. You're right, I did intend to use `tz = tz`.

PR #588, opened 2024-12-19
Initial separation of `tabyl`

### LINKED ISSUE #240: Breaking off tabyl and adorn_ into their own package?
There are a few open issues suggesting improvements to `adorn_` functions like having them use the `dplyr::select` helpers to specify columns.  I think those should be addressed as part of a broader rethinking of the tabyl API.  And once that's on the table, I've come to agree that tabyl and its friends belong in their own package.  janitor can be purely for data cleaning and focus on new functions like #50. 

Note: I have seen dreams discussed of a "grammar of tables" package, I hope _someone_ builds that package but `tabyl` will still just focus on counts and things you do to tables of counts.

Before making a new repo and going down that path: what do people think?  What's most weighing on my mind is, if someone wants to load janitor and <new pkg> together, would we need to rename tabyl and its friends to avoid namespace conflicts?  Naming things is hard 😩

billdenney:
Fix #240

This PR should import all functions from `tabyl` before being committed. That will require `tabyl` being on CRAN.


sfirke:
Nicely done! So first order of business is, get `tabyl` onto CRAN? They are slower with new package reviews than they are with updates.

We are past due to fix #582 which was closed in #584. I wonder if I should just branch off the repo at 2.2.0 and cherry that one fix in, then submit to CRAN as 2.2.1. That should get a very fast review and maybe beat the holiday rush?

billdenney:
Yeah, the first order of business is to get tabyl onto CRAN. I agree that this may be better after the janitor update, but if we are lucky, they may review tabyl quickly... Maybe?

I think that we nearly have tabyl to a good state to be a separate package. Maybe try to push it when everything is cleaned up there and see if it (surprisingly) goes quickly?

billdenney:
@sfirke, If some of the author(s) of `janitor` were solely focused on `tabyl` and not other `janitor` functions, they should probably be removed with this merge (because they will be in the `tabyl` list). Can you please review that?

billdenney:
Before we re-import the `tabyl` functions and after `tabyl` is on CRAN, I will test reverse dependencies and notify packages that depend on `janitor` to switch to `tabyl`.

I think that this means we will not get this PR into `janitor` before the required CRAN update. (Let's give ourselves breathing room.)

